PR #4, opened 2019-04-09
Rewrite configuration.py using Document class

ppessi:


apluslms_roman/configuration.py
```diff
@@ -56,36 +60,11 @@ def find_from(cls, path, name=None, prefixes=None):
...
-            raise CourseConfigError("Configuration from {} is invalid".format(source or '-'))
-
-        for key, value in config.items():
-            if key[0] == '_':
-                raise CourseConfigError('Invalid option {} in configuration from {}'.format(key, source or '-'))
-            setattr(self, key, value)
+        return super().load(config, version=cls.version, allow_missing=False)
 
+    def steps(self):
+        return self._container._documents[0][1]["steps"]
```
raphendyr: `self['steps']` should work here? In addition, steps could be a property.

apluslms_roman/builder.py
```diff
@@ -10,16 +10,16 @@
 
 class Builder:
     def __init__(self, engine, config, observer=None):
-        assert isdir(config.__path__), "Course configuration path doesn't exists"
+        assert isdir(config._container._dir), "Course configuration path doesn't exist"
         self.config = config
-        self.path = config.__path__
+        self.path = config._container._dir
```
raphendyr: I think path/dir should be a property of `config`. e.g `self.path = config.path`

apluslms_roman/builder.py
```diff
@@ -10,16 +10,16 @@
 
 class Builder:
     def __init__(self, engine, config, observer=None):
-        assert isdir(config.__path__), "Course configuration path doesn't exists"
+        assert isdir(config._container._dir), "Course configuration path doesn't exist"
```
raphendyr: Can you change this from assert to generic raise some exception. I learned that asserts can be skipped with python optimization flag.

apluslms_roman/configuration.py
raphendyr: Actually, this could probably be a function, though maybe renamed to `get_steps`. Later, filters can be added to it (upcoming feature).

So, later we could use it like `steps = config.get_steps([0, 2, 4, "rst"])` for example.

raphendyr: Hmm. There still could be a property `steps` that returns all steps.

apluslms_roman/builder.py
```diff
@@ -10,16 +10,17 @@
 
 class Builder:
     def __init__(self, engine, config, observer=None):
-        assert isdir(config._container._dir), "Course configuration path doesn't exist"
+        if not isdir(config.dir):
+            raise ValueError("config.__path__ isn't a directory.")
```
raphendyr: Need to update the error message

apluslms_roman/configuration.py
```diff
@@ -51,41 +60,17 @@ def find_from(cls, path, name=None, prefixes=None):
                     break
         else:
             if is_legacy_configuration(path):
-                return cls(LEGACY_CONFIG,
-                           path=path,
-                           source=LEGACY_SOURCE)
+                print("legacy")
```
raphendyr: debug print is left in the commit

apluslms_roman/settings.py
```diff
@@ -47,7 +47,6 @@ def get_config_path(cls, filename=None):
     def load(cls, path, allow_missing=False):
         return super().load(path, version=cls.version, allow_missing=allow_missing)
 
-
```
raphendyr: This is probably here by accident?

ppessi: what exactly should be done about it? another commit where the line is added back or rebase it so that the commit isn't made..?

raphendyr: You can rebase (fixup / squash) to remove the change.

apluslms_roman/backends/__init__.py
```diff
@@ -26,21 +27,31 @@ class BuildStep:
...
+                index,
+                clean_image_name(data['img']),
+                data.get('cmd'),
+                data.get('mnt'),
+                dict(data.get('env', {})),
+                data.get('name'),
+            )
         else:
-            return cls(clean_image_name(data), None, None, None)
+            return cls(0, clean_image_name(data))
```
raphendyr: Probably `index`?

apluslms_roman/cli.py
```diff
@@ -305,13 +326,23 @@ def build_action(context):
...
 
+    builder = engine.create_builder(config)
+
+    # TODO: option to print more?
+    if (context.parser.parse_args().steps):
+        steps = builder.get_steps()
+        num_len = len(str(len(steps)))
+        name_len = len(max(steps, key=lambda s: len(s.name or "")).name)
+        for step in steps:
+            step_str = "{:%dd}. {:%ds} {}" % (num_len, name_len)
```
raphendyr: No need recalculate for every iteration (no changes to data per item in for loop). So, can be moved before for.

apluslms_roman/configuration.py
```diff
@@ -20,29 +22,31 @@
...
+    name = 'course_config'
+    schema = name
+    version = Version(2, 0)
 
-class CourseConfig:
     @classmethod
     def find_from(cls, path, name=None, prefixes=None):
-        if not name:
-            name = DEFAULT_NAME
+        name = DEFAULT_NAME
```
raphendyr: Not sure this changes is supposed to be here. Name argument doesn't do anything now?

apluslms_roman/configuration.py
```diff
@@ -20,29 +22,31 @@
...
 ))
 
-
 class CourseConfigError(Exception):
     pass
 
-
 def is_legacy_configuration(path):
-    """Checks if path is legacy a-plus-rst-tools course"""
+    # Checks if path is legacy a-plus-rst-tools course
```
raphendyr: Why is the doc string changed to a comment? 🤔

apluslms_roman/builder.py
```diff
@@ -10,18 +11,33 @@
 
 class Builder:
     def __init__(self, engine, config, observer=None):
-        assert isdir(config.__path__), "Course configuration path doesn't exists"
+        # if not isdir(config.__path__):
```
raphendyr: This comment can be removed.

apluslms_roman/builder.py
```diff
@@ -10,18 +11,33 @@
...
+
+    def get_steps(self, refs: list = None):
+        steps = [BuildStep.from_config(i, step) for i, step in enumerate(self.config.steps)]
+        if refs:
+            name_dict = {step.name: step for step in steps}
+            step_list = [int(ref) if ref.isdigit() else ref for ref in refs]
+            try:
+                steps = [steps[ref] if isinstance(ref, int) else name_dict[ref] for ref in step_list]
+            except KeyError as e:
+                raise ValueError("No step named " + str(e))
```
raphendyr: These strings should be wrapped with `_(...)`, so when translations are introduced, they are automatically included. So, `_("No step named '{}'").format(e)`.

apluslms_roman/cli.py
```diff
@@ -197,6 +212,12 @@ def add_cli_actions(parser):
         help=_("build the course (default action)"))
     build.add_argument('course', nargs='?',
         help=_("location of the course definition (default: current working dir)"))
+    build.add_argument('--buildsteps', nargs='+',
```
raphendyr: This should include `-s` and long option should be `--steps`. Then in `build_action` we can check if first item of `steps` is a `?`. In addition, build subcommand could have `--list-steps` flag to do the same.

apluslms_roman/cli.py
```diff
@@ -305,13 +326,23 @@ def build_action(context):
     config = get_config(context)
     engine = get_engine(context)
 
+    builder = engine.create_builder(config)
+
+    # TODO: option to print more?
+    if (context.parser.parse_args().steps):
```
raphendyr: No need to re-parse arguments. `context.args.steps`.

So, if steps can be listed with option `--list-steps` or `--steps=?`, then this would become, `if context.args.list_steps or (context.args.steps and context.args.steps[0] == '?':`.

In addition, TODO is implemented here?

apluslms_roman/cli.py
```diff
@@ -305,13 +326,23 @@ def build_action(context):
...
+            print(step_str.format(step.ref, step.name or "", step.img))
+        return
+
     if not verify_engine(engine, only_when_error=True):
         return 1
 
     # build course
-    builder = engine.create_builder(config)
-    result = builder.build()
-    print(result)
```
raphendyr: `result` is no more printed?

apluslms_roman/cli.py
```diff
@@ -305,13 +326,23 @@ def build_action(context):
...
+    result = builder.build(context.parser.parse_args().buildsteps)
```
raphendyr: list of steps could be prepared using code `steps = list(chain.from_iterable(x.split(',') for x in context.args.steps))` (chain is from `itertools`). This way user can also say `--steps=foo,bar` instead of just `-s foo -s bar`, but both would be supported.

raphendyr:
You should rebase this on top of target branch. I.e. `git rebase upstream/milestone0.3` instead of just `git rebase -i` for example. That will drop my commit from the PR.

apluslms_roman/configuration.py
ppessi: hm probably messed something up during rebasing

apluslms_roman/configuration.py
ppessi: I thought it was unnecessary since it's a one-line comment, is there some other reason to use strings?

ppessi: okay googled it, changing it back

apluslms_roman/cli.py
ppessi: what I meant by "option to print more" is option to print more data of the steps, currently it's just the img.

apluslms_roman/cli.py
ppessi: I probably thought it was a debug print I left there.

apluslms_roman/cli.py
ppessi: context.args.list_steps doesn't work. list-steps currently:
`parser.add_argument('-ls', '--list-steps', action='store_true')`

ppessi: at least in zsh steps=? doesn't work, the ? needs to be quoted.

apluslms_roman/cli.py
ppessi: `-s foo -s bar` doesn't work, it only builds bar. `-s foo bar` or `--steps foo bar` work, do we still want to support `--steps=foo,bar` too?

apluslms_roman/cli.py
```diff
@@ -210,6 +212,11 @@ def add_cli_actions(parser):
         help=_("build the course (default action)"))
     build.add_argument('course', nargs='?',
         help=_("location of the course definition (default: current working dir)"))
+    build.add_argument('-s', '--steps', nargs='+',
+        help=_("select which steps to build and in which order (use either index or step name)"))
+    build.add_argument('-ls', '--list-steps',
```
raphendyr: Double char short options are always problematic. What would `-ls foo` mean, is it `-l --steps foo` or `--list-steps foo`. Short options can typically be grouped when given to commands and this is behavior what could be expected by the user. I don't personally see need for a short option, as the `-s ?` is quite intuitive too (I have seen that some times, best I could find now is `kvm -device ?`, which includes bad example of an argument itself).

apluslms_roman/cli.py
```diff
@@ -210,6 +212,11 @@ def add_cli_actions(parser):
         help=_("build the course (default action)"))
     build.add_argument('course', nargs='?',
         help=_("location of the course definition (default: current working dir)"))
+    build.add_argument('-s', '--steps', nargs='+',
```
raphendyr, line 238: I'm not yet sure if `nargs='+'` or `action='append'`. Though, because we have positional argument `course`, then command `roman build -s foo bar some_path` becomes obscure and would require format `roman build some_path -s foo bar`, which would be relatively ok. I'm not sure eactly how ArgumentParser handles this case internally. Biggest concern is that `roman build -s foo bar` is parsed as `{'steps': 'foo', 'course': 'bar'}`.

Thus, safer option would be the `action='append'` and would then require format `roman build -s foo -s bar some_path` witch is overly verbose and is reason why I was looking for option `roman build -s foo,bar some_path` or if I want to be verbose `roman build --steps=foo,bar some_path`.

apluslms_roman/cli.py
raphendyr: First time I actually where thinking that `--steps` used `action='append'` instead of `nargs='+'`, thus I thought `-s foo -s bar` would be the syntax. My bad. Any case, more comments on the topic on the new version of the file.

apluslms_roman/cli.py
ppessi: Well in zsh you need to quote the ? which makes it rather hard imo. I'll make "-l" the short option.

apluslms_roman/cli.py
raphendyr: I removed the `course` parameter in current milestone0.3, so the `nargs='+'` is now the better alternative. Still, it would be nice to be able to write `--steps=foo,bar` too.

apluslms_roman/cli.py
raphendyr: Yep, `?` is converted to a single character, if there exists a single character file. Shells typically pass that as is if it won't expand. E.g.

```
% echo ?; touch a; echo ?
?
a
```
Thus, `?` isn't very good, but still relatively acceptable alternative as there typically shouldn't be files with only a single character.

ppessi: should "roman -s ?" work or is it ok if only `roman -l` and `roman build -s ?` work?

apluslms_roman/cli.py
ppessi: I'm making the assumption that step names don't have commas in them

apluslms_roman/cli.py
ppessi: chain doesn't work, used an if clause instead

apluslms_roman/cli.py
raphendyr: Only build related command is`roman`. If user needs to add any options, then he is forced to write `roman build <something>`. Of course `roman -f foobar.yml` will work too, but yeah, rest is via the subcommand.

raphendyr:
Manually merged work done under the PR in 3acf2635f9504cfd9bf3ec04bcc1617667f8990c

PR #5, opened 2019-04-25
Step names

ppessi:
The user can now select which steps to build.


apluslms_roman/builder.py
```diff
@@ -10,18 +10,24 @@
...
+        return [BuildStep.from_config(i, step) for i, step in enumerate(self.config.steps)]
+
+    def build(self, step_list=[]):
         backend = self._engine.backend
         observer = self._observer
-        steps = [BuildStep.from_config(step) for step in self.config.steps]
+        steps = self.buildSteps()
+        if (step_list):
+            name_dict = {step.name: step for step in steps}
+            steps = [name_dict[step] if step in name_dict.keys() else steps[int(step)] for step in step_list]
```
raphendyr: This probably should first check if the `step` (probably should be called `ref` here) is number and convert it to integer and then we can raise correct errors.

e.g.
```
step_list = [int(ref) of ref.isdigit() else ref for ref in step_list]
try:
  steps = [steps[ref] if isinstance(ref, int) else name_dict[ref] for ref in step_list]
except KeyError as e:
  # no step with name ref
except IndexError as e:
  # ref is out of step list
```

Finally, we probably should check if there is duplicates. Probably should log a warning that some step is multiple times. Or we could just drop duplicates. `OrderedDict.from_keys(steps).keys()` will yield all steps only once in same order.

apluslms_roman/builder.py
```diff
@@ -10,18 +10,24 @@
...
-        self.path = config.__path__
+        self.path = config.dir
         self._engine = engine
         self._observer = observer or StreamObserver()
 
-    def build(self):
+    def buildSteps(self):
+        return [BuildStep.from_config(i, step) for i, step in enumerate(self.config.steps)]
+
+    def build(self, step_list=[]):
```
raphendyr: `, step_list=None` is bit more nice. If you like to tell the type, then we can write `, step_list: list = None`. Name `step_refs` might be a bit better.

apluslms_roman/builder.py
```diff
@@ -10,18 +10,24 @@
...
+        if not isdir(config.dir):
+            raise ValueError("config.dir isn't a directory.")
         self.config = config
-        self.path = config.__path__
+        self.path = config.dir
         self._engine = engine
         self._observer = observer or StreamObserver()
 
-    def build(self):
+    def buildSteps(self):
```
raphendyr: Typically we use underscores here (`build_steps`). In addition, as this is "getter", we can use property again, though, you could add `step_list` / `step_refs` as parameter. Thus, this probably should be `def get_steps(refs)`.

apluslms_roman/builder.py
```diff
@@ -10,18 +10,24 @@
...
+        return [BuildStep.from_config(i, step) for i, step in enumerate(self.config.steps)]
+
+    def build(self, step_list=[]):
         backend = self._engine.backend
         observer = self._observer
-        steps = [BuildStep.from_config(step) for step in self.config.steps]
+        steps = self.buildSteps()
+        if (step_list):
```
raphendyr: no brackets needed in python :)

apluslms_roman/cli.py
```diff
@@ -139,7 +139,11 @@ def create_parser(version=__version__,
         action='version',
         version="%%(prog)s %s" % (version,),
         help=_("print a version info and exit"))
-
+    parser.add_argument('-s', '--steps',
```
raphendyr, line 158: These should be set only for the `build` sub action, so define them for `build`parser here: https://github.com/apluslms/roman/blob/5d77d5391bc6b657aa25b98892aff08673ba153b/apluslms_roman/cli.py#L199

apluslms_roman/cli.py
```diff
@@ -305,12 +309,27 @@ def build_action(context):
     config = get_config(context)
     engine = get_engine(context)
 
+    builder = engine.create_builder(config)
+
+    # TODO: option to print more?
+    if (context.parser.parse_args().steps):
+        steps = builder.buildSteps()
+        num_len = len(str(len(steps)))
+        name_len = len(max(steps, key=lambda s: len(s.name or "")).name)
```
raphendyr, line 335: You can build string format string here. Something in lines of
```
fmt = "{:%dd}. {:%ds} {}" % (num_len, name_len)
fmt.format(step.ref, step.name or "", step.img)
```

Mixing two formatting methods is fine here to keep the format string generation simple (no need for escapes)

apluslms_roman/backends/__init__.py
```diff
@@ -26,21 +27,32 @@ class BuildStep:
...
+                data.get('name'),
+                index
+            )
         else:
-            return cls(clean_image_name(data), None, None, None)
+            return cls(clean_image_name(data), None, None, None, None)
 
-    def __init__(self, img, cmd, mnt, env):
-        self.img, self.cmd, self.mnt, self.env = img, cmd, mnt, env
+    def __init__(self, img, cmd, mnt, env, name, ref):
```
raphendyr: We could reorder these to be `ref, img, cmd=None, mnt=None, env=None,, name=None`.
And when setting `self.name` we could default that to `ref` if no name is provided. (`self.name = name or ref`)

apluslms_roman/backends/__init__.py
```diff
@@ -26,21 +27,32 @@ class BuildStep:
...
+                img,
+                data.get('cmd'),
+                data.get('mnt'),
+                dict(data.get('env', {})),
+                data.get('name'),
+                index
+            )
         else:
-            return cls(clean_image_name(data), None, None, None)
+            return cls(clean_image_name(data), None, None, None, None)
```
raphendyr: with the change to `__init__` this would become `return cls(index, clean_image_name(data))`

apluslms_roman/backends/__init__.py
```diff
@@ -26,21 +27,32 @@ class BuildStep:
...
-    def from_config(cls, data):
-        if isinstance(data, dict):
+    def from_config(cls, index, data):
+        if isinstance(data, Mapping):
             if 'img' not in data:
                 raise RuntimeError("Missing image name (img) in step configuration: {}".format(data))
             img = clean_image_name(data['img'])
-            return cls(img, data.get('cmd'), data.get('mnt'), data.get('env'))
+            return cls(
+                img,
```
raphendyr: `clean_image_name(data['img'])` can be moved here as `cls(...)` is now on multiple lines,

apluslms_roman/cli.py
raphendyr: p.s. format spec: https://docs.python.org/3.4/library/string.html#formatspec

apluslms_roman/builder.py
ppessi: AttributeError: type object 'collections.OrderedDict' has no attribute 'from_keys'

apluslms_roman/builder.py
ppessi: this is why I shouldn't code with other languages... I keep trying to call functions without brackets because of scala

apluslms_roman/cli.py
ppessi: do you mean both --steps and --buildsteps? why should --steps be added to build?

apluslms_roman/backends/__init__.py
ppessi: currently when listing steps I don't list a step name if step is none. if name is ref then that doesn't work and the only way to check if the name is ref is to compare it with the ref (or check if name is int).

apluslms_roman/cli.py
ppessi: if I move --buildsteps to build I can't get the values from context.parser.parse_args(). where can I find the buildsteps? (at least I think it was this that broke it)

apluslms_roman/builder.py
raphendyr: Forgot that with these types there isn't separator, so `list(OrderedDict.fromkeys(steps))`.  `.keys()` is not actually required as iterator over dictionary does the same and we need the `list` there with `keys()` too.

For reference, conversation about this topic: https://stackoverflow.com/questions/480214/how-do-you-remove-duplicates-from-a-list-whilst-preserving-order

apluslms_roman/cli.py
raphendyr: I fixed how base parser defaults are set when no subcommand is given. So, this should work nicely now. You need to rebase your changes on top of current milestone0.3

apluslms_roman/backends/__init__.py
raphendyr: Good point. Guess it's better have name as None when not set. `__str__` can then use name or ref if there is no name.

PR #8, opened 2019-05-28
Allow setting host path with command line arguement, with environment variable and with roman config

JohnDing1995:
Tell docker back-end the location of source file on host machine(if running  roman on a container)
The priority order is 
1. Global configure
2. Environment variable
3. Command line parament

### DIFF
--- a/apluslms_roman/backends/__init__.py
+++ b/apluslms_roman/backends/__init__.py
@@ -1,8 +1,11 @@
+import logging
 from collections import namedtuple
 from collections.abc import Mapping
+from pathlib import PurePosixPath
 
 from ..observer import BuildObserver
 
+logger = logging.getLogger(__name__)
 
 BACKENDS = {
     'docker': 'apluslms_roman.backends.docker.DockerBackend',
@@ -112,3 +115,18 @@ class Backend:
 
     def version_info(self):
         pass
+
+    def get_host_path(self, original):
+        mapping = self.environment.environ.get('directory_map', {})
+        if not mapping:
+            return original
+        logger.debug("Get mapping from environment:%s", mapping)
+        path = PurePosixPath(original)
+        for container, host in mapping.items():
+            try:
+                logger.debug("Mapping:%s:%s", container, host)
+                return str(PurePosixPath(host, path.relative_to(container)))
+            except ValueError:
+                logger.error("Error when composing new path!")
+                continue
+        return ValueError("Unable to map path '%s' to any backend host path" % (path))
--- a/apluslms_roman/backends/docker.py
+++ b/apluslms_roman/backends/docker.py
@@ -1,6 +1,7 @@
-import docker
-from os.path import join
+import logging
+from os.path import join, expanduser
 
+import docker
 from apluslms_yamlidator.utils.decorator import cached_property
 
 from ..utils.translation import _
@@ -12,6 +13,8 @@ from . import (
 
 Mount = docker.types.Mount
 
+logger = logging.getLogger(__name__)
+
 
 class DockerBackend(Backend):
     name = 'docker'
@@ -22,14 +25,28 @@ You might be able to add yourself to that group with 'sudo adduser docker'.""")
     @cached_property
     def _client(self):
         env = self.environment.environ
-        kwargs = {}
-        version = env.get('DOCKER_VERSION', None)
-        if version:
-            kwargs['version'] = version
-        timeout = env.get('DOCKER_TIMEOUT', None)
-        if timeout:
-            kwargs['timeout'] = timeout
-        return docker.from_env(environment=env, **kwargs)
+        params = {
+            'base_url': env.get('host'),
+            'version': env.get('version'),
+        }
+        if 'timeout' in env:
+            params['timeout'] = env['timeout']
+
+        # false values: 0, false, '', unset
+        # true values: 1, true, "yes"
+        tls_verify = bool(env.get('tls_verify', False))
+        cert_path = env.get('cert_path') or None
+        if tls_verify or cert_path:
+            if not cert_path:
+                cert_path = join(expanduser('~'), '.docker')
+            params['tls'] = docker.tls.TLSConfig(
+                client_cert=(join(cert_path, 'cert.pem'), join(cert_path, 'key.pem')),
+                ca_cert=join(cert_path, 'ca.pem'),
+                verify=tls_verify,
+                ssl_version=env.get('tls_ssl_version'),
+                assert_hostname=tls_verify and env.get('tls_assert_hostname'),
+            )
+        return docker.DockerClient(**params)
 
     def _run_opts(self, task, step):
         env = self.environment
@@ -41,16 +58,19 @@ You might be able to add yourself to that group with 'sudo adduser docker'.""")
             user='{}:{}'.format(env.uid, env.gid),
         )
 
+        path = self.get_host_path(task.path)
+
+        logger.debug("Final path is:%s", path)
         # mounts and workdir
         if step.mnt:
-            opts['mounts'] = [Mount(step.mnt, task.path, type='bind', read_only=False)]
+            opts['mounts'] = [Mount(step.mnt, path, type='bind', read_only=False)]
             opts['working_dir'] = step.mnt
         else:
             wpath = self.WORK_PATH
             opts['mounts'] = [
                 Mount(wpath, None, type='tmpfs', read_only=False, tmpfs_size=self.WORK_SIZE),
-                Mount(join(wpath, 'src'), task.path, type='bind', read_only=True),
-                Mount(join(wpath, 'build'), join(task.path, '_build'), type='bind', read_only=False),
+                Mount(join(wpath, 'src'), path, type='bind', read_only=True),
+                Mount(join(wpath, 'build'), join(path, '_build'), type='bind', read_only=False),
             ]
             opts['working_dir'] = wpath
 
--- a/apluslms_roman/builder.py
+++ b/apluslms_roman/builder.py
@@ -9,6 +9,7 @@ from .observer import StreamObserver
 from .utils.importing import import_string
 from .utils.translation import _
 
+
 class Builder:
     def __init__(self, engine, config, observer=None):
         if not isdir(config.dir):
@@ -21,7 +22,7 @@ class Builder:
 
     def get_steps(self, refs: list = None):
         steps = [BuildStep.from_config(i, step)
-            for i, step in enumerate(self.config.steps)]
+                 for i, step in enumerate(self.config.steps)]
         if refs:
             name_dict = {step.name: step for step in steps}
             refs = [int(ref) if ref.isdigit() else ref.lower() for ref in refs]
@@ -64,11 +65,9 @@ class Engine:
 
         name = getattr(backend_class, 'name', None) or backend_class.__name__.lower()
         env_prefix = name.upper() + '_'
-        env = {k: v for k, v in environ.items() if k.startswith(env_prefix)}
+        env = {key: value for key, value in environ.items() if key.startswith(env_prefix)}
         if settings:
-            for k, v in settings.get(name, {}).items():
-                if v is not None and v != '':
-                    env[env_prefix + k.replace('-', '_').upper()] = v
+            env.update(settings.get(name, {}))
         self._environment = Environment(getuid(), getegid(), env)
 
     @cached_property
--- a/apluslms_roman/schemas/roman_settings-v1.0.yaml
+++ b/apluslms_roman/schemas/roman_settings-v1.0.yaml
@@ -24,6 +24,10 @@ properties:
     type: object
     additionalProperties: false
     properties:
+      directory_map:
+        title: docker container-host machine path mapping
+        description: The dictornary mapping between docker and it's host
+        type: object
       host:
         title: docker host
         description: the URL to the Docker host
--- a/apluslms_roman/utils/path_mapping.py
+++ b/apluslms_roman/utils/path_mapping.py
@@ -1,0 +1,25 @@
+import json
+import logging
+import re
+from os import environ
+from apluslms_yamlidator.document import find_ml
+
+logger = logging.getLogger(__name__)
+json_re = re.compile(r'^(?:["[{]|(?:-?[1-9]\d*(?:\.\d+)?|null|true|false)$)')
+
+
+def nest_dict(flat_dict):
+    nested = {}
+    for keys, value in flat_dict.items():
+        dict_, key = find_ml(nested, keys, create_dicts=True)
+        dict_[key] = value
+    return nested
+
+
+def load_from_env(env_prefix=None, decode_json=True):
+    if decode_json:
+        decode = lambda s: json.loads(s) if json_re.match(s) is not None else s
+    else:
+        decode = lambda s: s
+    env = {key[len(env_prefix):].lower(): decode(value) for key, value in environ.items() if key.startswith(env_prefix)}
+    return nest_dict(env)
--- a/tests/utils/test_path_mapping.py
+++ b/tests/utils/test_path_mapping.py
@@ -1,0 +1,64 @@
+import os
+import unittest
+from json import loads
+
+from apluslms_roman.utils.path_mapping import json_re, load_from_env
+
+test_case_loadable = (
+    'true',
+    'false',
+    'null',
+    '123',
+    '-123',
+    '3.14',
+    '-3.14',
+    '{"foo": "bar"}',
+    '[1, 2, 3]',
+    '"foo bar"'
+)
+
+test_case_not_loadable = (
+    "/foobar.py",
+    "text",
+    "yes",
+    "0123123",
+)
+
+
+class TestJsonLoadable(unittest.TestCase):
+
+    def test_loadable_not_raise(self):
+        for case in test_case_loadable:
+            with self.subTest(non_json=case):
+                loads(case)
+
+    def test_not_loadable_raise(self):
+        for case in test_case_not_loadable:
+            with self.subTest(non_json=case):
+                with self.assertRaises(ValueError, msg="Testing:{}".format(case)):
+                    loads(case)
+
+
+class TestJsonRegex(unittest.TestCase):
+
+    def test_loadable_match(self):
+        for case in test_case_loadable:
+            with self.subTest(non_json=case):
+                self.assertTrue(json_re.match(case)is not None, msg="Testing:{}".format(case))
+
+    def test_not_loadable_not_match(self):
+        for case in test_case_not_loadable:
+            with self.subTest(non_json=case):
+                self.assertFalse(json_re.match(case) is not None, msg="Testing:{}".format(case))
+
+
+class TestLoadFromEnv(unittest.TestCase):
+
+    def test_with_decode_json(self):
+        os.environ['DOCKER.FOO.BAR'] = '123'
+        self.assertEqual({'foo': {'bar': '123'}}, load_from_env('DOCKER.', False))
+
+    def test_without_decode_json(self):
+        os.environ['DOCKER.FOO.BAR'] = '123'
+        self.assertEqual({'foo': {'bar': 123}}, load_from_env('DOCKER.', True))
+


apluslms_roman/utils/path_mapping.py
```diff
@@ -0,0 +1,30 @@
...
+            ret = PurePosixPath(v).joinpath(relative_path)
+            return str(ret)
+        except ValueError:
+            print("Error when composing new path!")
+            return -1
+    return str(ret)
+
+
+def get_pair_form_env(key, json_str):
+    if key == "DOCKER_DIRECTORY_MAP":
```
raphendyr: Utility functions should be generic, like functions in standard library. They take input and create output and don't have special cases. This function does something only, when it's argument has specific value. E.g. this if statement should be where this function is called.

Basically, utility function should be testable with it's own unit tests.

apluslms_roman/utils/path_mapping.py
```diff
@@ -0,0 +1,30 @@
...
+        try:
+            ret = json.loads(json_str)
+            if isinstance(ret, dict):
+                return ret
+        except json.decoder.JSONDecodeError:
+            print("Error, check your json string")
+            return json_str
+    else:
```
raphendyr, line 22: This function has 4 return states, where one is `return None` (implicit end of function). We get there when json doesn't return a dictionary (e.g. when `json_str` has value `true`). If this `else` is removed, then this function would only need 2 return statements `return ret` and `return json_str`.

apluslms_roman/utils/path_mapping.py
```diff
@@ -0,0 +1,30 @@
...
+    orig_path = PurePosixPath(original)
+    for k, v in mapping.items():
+        try:
+            print(k, v)
+            relative_path = orig_path.relative_to(k)
+            ret = PurePosixPath(v).joinpath(relative_path)
+            return str(ret)
+        except ValueError:
+            print("Error when composing new path!")
+            return -1
```
raphendyr: I presume the for loop should continue, so maybe `pass` or `continue´.

apluslms_roman/utils/path_mapping.py
```diff
@@ -0,0 +1,30 @@
...
+    ret = original
+    orig_path = PurePosixPath(original)
+    for k, v in mapping.items():
+        try:
+            print(k, v)
+            relative_path = orig_path.relative_to(k)
+            ret = PurePosixPath(v).joinpath(relative_path)
+            return str(ret)
+        except ValueError:
+            print("Error when composing new path!")
```
raphendyr: There shouldn't be `print` statements in any other file than `cli.py`. Logging can be used, when the user doesn't need to see the messages, but the messages could help him in some cases (e.g. to debug).

apluslms_roman/roman_config,yml
```diff
@@ -0,0 +1,7 @@
+
```
raphendyr: This file shouldn't be in the repo.

apluslms_roman/cli.py
```diff
@@ -172,6 +172,10 @@ def create_parser(version=__version__,
         dest='project_config',
         metavar=_('FILE'),
         help=_("use the FILE as the project configuration file"))
+    parser.add_argument('-p', '--path',
```
raphendyr: let's omit this argument for now and let's add it later if needed.

apluslms_roman/utils/path_mapping.py
```diff
@@ -0,0 +1,30 @@
...
+            return -1
+    return str(ret)
+
+
+def get_pair_form_env(key, json_str):
```
raphendyr: This function could be refactored to something called `def load_from_environ(prefix: str) -> dict:`. It would:

1. read all values from `os.environ`, where key starts with `prefix`
2. checks if the value matches re `r'^(?:["[{]|(?:\d+|null|true|false)$)', if so then json decode it
3. if decode fails, raise error
4. lowercase the key
5. if there is `.` (dot) in the key, then split the key to parts (e.g. `"FOO_BAR.BAZ"` -> `["FOO_BAR", "BAZ"]`)
6. build result nested dictionaries from the data, presuming that missing elements are dicts

For example:
```
os.environ = {
  'PREFIX_FOO.BAR': 'true',
  'PREFIX_FOO.BAZ': '[1, 2, 3]',
}
load_from_environ('PREFIX_') == {
  'foo': {
    'bar': True,
    'baz': [1, 2, 3],
  }
}
```

apluslms_roman/builder.py
```diff
@@ -60,11 +88,13 @@ def __init__(self, backend_class=None, settings=None):
 
         name = getattr(backend_class, 'name', None) or backend_class.__name__.lower()
         env_prefix = name.upper() + '_'
-        env = {k: v for k, v in environ.items() if k.startswith(env_prefix)}
+        env = {k:  get_pair_form_env(k, v) for k, v in environ.items() if k.startswith(env_prefix)}
```
raphendyr: this and and few lines below, could be rewritten to use `load_from_environ` defined in different comment. Thus, create a dictionary, which can be then passed as it to backend class. Environment dictionary can then be converted a list of environment variables if really needed.

apluslms_roman/builder.py
```diff
@@ -32,11 +34,35 @@ def get_steps(self, refs: list = None):
...
+            # Using path in global config file
+        print("dir mapping set in config:", self._engine._dir_mapping if hasattr(self._engine, '_dir_mapping') else None)
+        print("-p paramater set:", host_path)
+        if self._engine._environment.environ.get('DOCKER_DIRECTORY_MAP', None) is not None:
+            if hasattr(self._engine, '_dir_mapping'):
+                print("using config")
+            else:
+                print("Using environment variable")
+            mapping = self._engine._environment.environ.get('DOCKER_DIRECTORY_MAP', None)
+            path = get_host_path(self.path, dict(mapping))
```
raphendyr: This should be about the only thing that is needed here, I presume.

apluslms_roman/builder.py
```diff
@@ -4,10 +4,13 @@
 from apluslms_yamlidator.utils.decorator import cached_property
 from apluslms_yamlidator.utils.collections import OrderedDict
 
+from apluslms_roman.utils.path_mapping import get_host_path
```
raphendyr: You are importing with this absolute path and with a relative path below from the same module `path_mapping`.

apluslms_roman/backends/docker.py
```diff
@@ -29,6 +29,9 @@ def _client(self):
         timeout = env.get('DOCKER_TIMEOUT', None)
         if timeout:
             kwargs['timeout'] = timeout
+        # host_path = env.get('DOCKER_HOST_PATH', None)
```
raphendyr: typically, code which is commented out, is not added to the git.

apluslms_roman/backends/docker.py
```diff
@@ -29,6 +29,9 @@ def _client(self):
...
+        # if host_path:
+        #     print("Host source file path is", host_path)
         return docker.from_env(environment=env, **kwargs)
```
raphendyr: `docker.from_env` is handy function, but it might become easier to use another one, when we change `env` to be a python dictionary with non-uppercase keys. Alternatively, this is the place to build minimalistic environment for the `from_env` from those variables we have available to use here.

ref: docker-py from_env: https://github.com/docker/docker-py/blob/master/docker/client.py#L85

apluslms_roman/utils/path_mapping.py
JohnDing1995: I'll change to 
```
def get_pair_form_env(key, json_str, read_key="DOCKER_DIRECTORY_MAP"):
    if key == read_key:
```

apluslms_roman/utils/path_mapping.py
JohnDing1995: the usage of this `else` is when the key we iterating is not the key we need to read (like `DOCKER_DIRECTORY_MAP`), we will return the origin value of the string. if we don't have this else, the function may empty all value except the value of `DOCKER_DIRECTORY_MAP`

See the usage of the function:
`env = {k:  get_pair_form_env(k, v) for k, v in environ.items() if k.startswith(env_prefix)}`

apluslms_roman/utils/path_mapping.py
JohnDing1995: I was thinking to return -1 so that builder will know there is error in composing from environment variable, so it will trie to using command line parameter or default instead.
But I forget that we could have multiple pair, in this case we also need to try pairs behind the failed one. so using pass is better

apluslms_roman/utils/path_mapping.py
raphendyr: FYI: The default value for a keyword argument should nearly always be `None`. For int values zero might be ok. Using Python functions that have more complex defaults becomes easily complex.

for reference: https://graysonkoonce.com/always-use-none-for-default-args-in-python/

apluslms_roman/builder.py
JohnDing1995: You mean this should not be changed?

apluslms_roman/utils/path_mapping.py
JohnDing1995: >     5\. if there is `.` (dot) in the key, then split the key to parts (e.g. `"FOO_BAR.BAZ"` -> `["FOO_BAR", "BAZ"]`)

You mean in this case we have two keys now?

apluslms_roman/builder.py
raphendyr: I presume when you do the rest of the stuff, the code around this comment will get pretty clean and this seemed to be the only important part.

Though, the dict copy in `dict(mapping)` seems unnecessary.

apluslms_roman/utils/path_mapping.py
JohnDing1995: In such case the setting cannot overwrite the environment variable, the environment variable after reading setting is :
```
{'directory_map': {'foo': 'bar', 'foo1': 'bar1'},
 'DOCKER_DIRECTORY_MAP': ChangesDict(data={'/data/': '/var/lib/docker/volumes/aplus_data/_data/'}, default={})}
```
Which shows that two variable exist at same time, see slack chat for more deatil

.gitignore
```diff
@@ -8,3 +8,5 @@ setup.cfg
 /dist/
 /venv/
 *.swp
+/apluslms_roman/*.yml
```
raphendyr: This line shouldn't be here. There is no need to ignore yaml files in that path.

.gitignore
```diff
@@ -8,3 +8,5 @@ setup.cfg
...
+.idea/
```
raphendyr: "no new line at end of file" :)

test/test_json_regex.py
```diff
@@ -0,0 +1,55 @@
...
+        data = None
+    return data
+
+
+class TestJsonRegex(unittest.TestCase):
+
+    def test_loadable_match(self):
+        for i, case in enumerate(test_case_loadable):
+            with self.subTest(i=i):
+                self.assertTrue(json_re.match(dumps(case))is not None, msg="Testing:{}".format(case))
```
raphendyr: You need to test `env_value_to_dict` function, not local regexp in this file.

apluslms_roman/utils/path_mapping.py
```diff
@@ -0,0 +1,81 @@
...
+            logger.debug("Error, check your json string")
+    return json_str
+
+
+def env_process_key(key, prefix):
+    return
+
+
+def env_value_to_dict(json_str):
+    if re.match(r'^(?:["[{]|(?:-?[1-9]\d*(?:\.\d+)?|null|true|false)$)', json_str):
```
raphendyr: The reg exp should be precompiled, so it would be fast to use. Now we compile the regexp every time we run env_value_to_dict.

apluslms_roman/utils/path_mapping.py
```diff
@@ -0,0 +1,81 @@
...
+        try:
+            ret = json.loads(json_str)
+            if isinstance(ret, dict):
```
raphendyr: `true` is valid json and should be passed as `True`. Thus, this if statement is not needed.

apluslms_roman/utils/path_mapping.py
```diff
@@ -0,0 +1,81 @@
...
+            if root not in ret:
+                ret[root] = {}
+            ret[root][key_list[1]] = v
+        else:
+            ret[k] = v
+    return ret
+
+
+def load_from_env(env_prefix=None, sep=None):
+    env_prefix = "" if env_prefix is None else env_prefix
```
raphendyr: you could define support function here, something like:

```
naive_json_re = re.compile(r'^(?:["[{]|(?:-?[1-9]\d*(?:\.\d+)?|null|true|false)$)')

def load_from_env(prefix=None, sep=None, decode_json=True):
  if decode_json:
    decode = lambda s: json.loads(s) if naive_json_re.match(s) is not None else s
  else
    decode = lambda s: s
  env = {k[len(env_prefix):].lower(): decode(v) for k, v in environ.items() if k.startswith(env_prefix)}
  ...
```

apluslms_roman/cli.py
```diff
@@ -294,11 +294,8 @@ def main():
     exit(context.run())
 
 
-
```
raphendyr: I don't see any point to include only these whitespace changes to this file in this PR. You can `git checkout origin/milestone0.4 apluslms_roman/cli.py` to get the original file.

apluslms_roman/builder.py
```diff
@@ -60,11 +76,13 @@ def __init__(self, backend_class=None, settings=None):
...
         env_prefix = name.upper() + '_'
-        env = {k: v for k, v in environ.items() if k.startswith(env_prefix)}
+        env = load_from_env(env_prefix, '.')
+        print("env without setting:", env)
         if settings:
             for k, v in settings.get(name, {}).items():
                 if v is not None and v != '':
-                    env[env_prefix + k.replace('-', '_').upper()] = v
+                    env[k] = dict(v) if isinstance(v, ChangesDict) else v
+        print("env after setting:", env)
```
raphendyr: Thre shouldn't be any `print` statements in other files than `cli.py` in the library. You can find all places of print in the roman with `git grep -F print '*.py'` for reference.

Prints are of course food when developing, but they shouldn't exist in the PR.

apluslms_roman/backends/docker.py
```diff
@@ -29,7 +30,7 @@ def _client(self):
         timeout = env.get('DOCKER_TIMEOUT', None)
         if timeout:
             kwargs['timeout'] = timeout
-        return docker.from_env(environment=env, **kwargs)
+        return DockerClient(env.get('DOCKER_HOST', None), version, timeout)
```
raphendyr: I was expecting something like:

```
params = {
  'base_url': env.get('host'),
  'version': env.get('version'),
}
if 'timeout' in env:
  params['timeout'] = env['timeout']

# false values: 0, false, '', None
# true values: 1, true, "yes", unset
tls_verify = bool(env.get('tls_verify', True))
cert_path = env.get('cert_path') or None

if tls_verify or cert_path:
  if not cert_path:
    cert_path = os.path.join(os.path.expanduser('~'), '.docker')
  params['tls'] = docker.tls.TLSConfig(
        client_cert=(os.path.join(cert_path, 'cert.pem'), os.path.join(cert_path, 'key.pem')),
        ca_cert=os.path.join(cert_path, 'ca.pem'),
        verify=tls_verify,
        ssl_version=env.get('tls_ssl_version'),
        assert_hostname=tls_verify and env.get('tls_assert_hostname'),
  )

return DockerClient(**params)
```

apluslms_roman/builder.py
```diff
@@ -2,12 +2,13 @@
...
 
 from apluslms_yamlidator.utils.decorator import cached_property
-from apluslms_yamlidator.utils.collections import OrderedDict
+from apluslms_yamlidator.utils.collections import OrderedDict, ChangesDict
+from apluslms_roman.utils.path_mapping import get_host_path, load_from_env, get_pair_form_env
 
 from .backends import BACKENDS, BuildTask, BuildStep, Environment
 from .observer import StreamObserver
 from .utils.importing import import_string
-from .utils.translation import _
```
raphendyr: This is used in this file, so it' can't be removed.

apluslms_roman/builder.py
raphendyr: Actually, there is no need to add anything to this file abouf `directory_map` spicifically.

You could add

```
def remap_path(self, path):
  map_ = self.environment.get('directory_map', {})
  return get_host_path(path, map_)
```

To Backend in https://github.com/apluslms/roman/blob/072f95e2bafee900e3c5789bb9fc5fa653cabb53/apluslms_roman/backends/__init__.py#L84

Then for every path in `DockerBackend._run_opts()` you could call `self.remap_path(path)`. https://github.com/apluslms/roman/blob/072f95e2bafee900e3c5789bb9fc5fa653cabb53/apluslms_roman/backends/docker.py#L35

apluslms_roman/builder.py
```diff
@@ -50,6 +64,8 @@ def __init__(self, backend_class=None, settings=None):
         if backend_class is None:
             if settings and 'backend' in settings:
                 backend_class = settings['backend']
+                if 'docker' in settings and 'directory_map' in settings['docker']:
```
raphendyr: This should not be needed. In addition, this is not a very good way to implement this any way. ref: https://github.com/apluslms/roman/pull/8/files#r292131426

apluslms_roman/builder.py
```diff
@@ -60,11 +76,13 @@ def __init__(self, backend_class=None, settings=None):
...
         name = getattr(backend_class, 'name', None) or backend_class.__name__.lower()
         env_prefix = name.upper() + '_'
-        env = {k: v for k, v in environ.items() if k.startswith(env_prefix)}
+        env = load_from_env(env_prefix, '.')
+        print("env without setting:", env)
         if settings:
             for k, v in settings.get(name, {}).items():
                 if v is not None and v != '':
-                    env[env_prefix + k.replace('-', '_').upper()] = v
+                    env[k] = dict(v) if isinstance(v, ChangesDict) else v
```
raphendyr: Unwrapping `ChangesDict`, shouldn't be needed. I should implement "unwarp" method to `Changes` classes, so it could be used, when exporting values in bare format.

Any case, environment values should overwrite settings (superseding command line arguments, until that is actually fixed, but that can't be done here cleanly).

.gitignore
JohnDing1995: I was thinking to ignore roman_config.yml

apluslms_roman/utils/path_mapping.py
JohnDing1995: Is a good idea to assign a `lambda` expression? Seems not recommended by PEP
See: https://lintlyci.github.io/Flake8Rules/rules/E731.html

apluslms_roman/builder.py
JohnDing1995: Replace with logging.debug()

.gitignore
raphendyr: It doesn't belong in that folder in the first place :)

apluslms_roman/utils/path_mapping.py
raphendyr: It's not as it removes function name from the debug info (e.g. stack trace). Though, it should be fine for case like this.

Alternatively same can be written:

```
naive_json_re = re.compile(r'^(?:["[{]|(?:-?[1-9]\d*(?:\.\d+)?|null|true|false)$)')

def load_from_env(prefix=None, sep=None, decode_json=True):
  if decode_json:
    def decode(s):
      return json.loads(s) if naive_json_re.match(s) is not None else s
  else:
    def decode(s):
      return s
  env = {k[len(env_prefix):].lower(): decode(v) for k, v in environ.items() if k.startswith(env_prefix)}
  ...
```

test/test_json_regex.py
JohnDing1995: It's not possible to test if regex works by only testing `env_value_to_dict` function

JohnDing1995: It's not easy and straight forward to test if regex works by only testing `env_value_to_dict` function
Neither from the type value of return nor if the return value is same with our test case, if we still using test case like 
```
test_case_loadable = (
    True,
    False,
    None,
    123,
    -123,
    3.14,
    -3.14,
    {'foo': 'bar'},
    [1, 2, 3],
    "foo bar"
)

test_case_not_loadable = (
    "/foobar.py",
    "text",
    "yes",
    "0123123",
)
```
I think good idea is to import compiled regex  `json_re` from  `apluslms_roman.utils.path_mapping`, rather than import and test this function

JohnDing1995: For example, if we using test case ` "foo bar"`(match regex) an testcase `"yes"`(not match with regex), both will return a `string` and both will make the return of `env_value_to_dict` equal with the parameter of this function, which makes us cannot tell if the regex real works or not.

raphendyr: Hmm.. you can import the compiled reg exp and test that directly, but also test the env_value_to_dict function.

That said, with `{'TEST1': '"foo bar"', 'TEST2': "yes"}` (first case and the second), you will get back `"foo bar"` and `"yes"`, where first is not the same as the input (one set of quotes is removed).

JohnDing1995: Ok i see the point

tests/test_json_regex.py
```diff
@@ -0,0 +1,64 @@
+import unittest
```
raphendyr: I presume we would like to name it `tests/utils/test_path_mapping.py`, so that it matches the module structure inside the package and starts with test, which is the default pattern used by unittest library.

I added some tests as reference in 5b6098320cd45ba0dfe5f32fc3a1e0c6444100e1

tests/test_json_regex.py
```diff
@@ -0,0 +1,64 @@
...
+
+test_case_not_loadable = (
+    "/foobar.py",
+    "text",
+    "yes",
+    "0123123",
+)
+
+
+def try_loads_json(json_str):
```
raphendyr: I would prefer if this function would be removed and the code would use `self.assertRaises` instead, if that would be possible.

tests/test_json_regex.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def test_not_loadable_equal(self):
+        for i, case in enumerate(test_case_not_loadable):
+            with self.subTest(i=i):
+                self.assertNotEqual(try_loads_json(dumps(case)), -1, msg="Testing:{}".format(case))
+
+    def test_loadable_type(self):
+        for i, case in enumerate(test_case_loadable):
+            with self.subTest(i=i):
+                self.assertNotEqual(env_value_to_dict(dumps(case)), dumps(case), msg="Testing:{}".format(case))
```
raphendyr: Use different classes for different functions/features, so it's easier to notice when reading the file.

raphendyr:
Some commit drops execution bit from `setup.py`. That shouldn't happen. Are you using `git add -a`, `git add -u` or `git commit -a` by any change?

tests/test_json_regex.py
JohnDing1995: pulled

JohnDing1995:
> Some commit drops execution bit from `setup.py`. That shouldn't happen. Are you using `git add -a`, `git add -u` or `git commit -a` by any change?

I was using `git add . && git commit -m "msg"` when making commits.
I'll using `git commit -am "msg"` for next commit.

tests/test_json_regex.py
JohnDing1995: Why do we put test cases under `/utils`?

apluslms_roman/builder.py
```diff
@@ -60,11 +65,13 @@ def __init__(self, backend_class=None, settings=None):
...
         env_prefix = name.upper() + '_'
-        env = {k: v for k, v in environ.items() if k.startswith(env_prefix)}
+        env = load_from_env(env_prefix, '.')
+        logger.debug("env without reading global config:{}".format(env))
         if settings:
             for k, v in settings.get(name, {}).items():
                 if v is not None and v != '':
-                    env[env_prefix + k.replace('-', '_').upper()] = v
+                    env[k] = v
+        logger.debug("env after read from global config:{}".format(env))
```
raphendyr: Always use logging with syntax: `logger.debug("old style %s format string", env)`. This way the formatting is not done unless the log line is actually printed somewhere. Code above does always call string.format, also when debug logs are disabled.

apluslms_roman/builder.py
```diff
@@ -1,14 +1,20 @@
...
+from apluslms_yamlidator.utils.decorator import cached_property
 
+from apluslms_roman.utils.path_mapping import load_from_env
 from .backends import BACKENDS, BuildTask, BuildStep, Environment
 from .observer import StreamObserver
 from .utils.importing import import_string
 from .utils.translation import _
 
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
```
raphendyr: Logging is configured by CLI or GUI, so don't ever touch logging level per file. With roman, you can add `-v` arguments to raise logging. `roman -vv` should result logging.DEBUG.

JohnDing1995: -v is already taken, maybe `-l` or `-ll`?

JohnDing1995: Circular import will happens if we try to import something form `cil`, so I'm thinking to put logger level to `env`

JohnDing1995: newest changes pused

apluslms_roman/utils/path_mapping.py
```diff
@@ -5,8 +5,7 @@
 from os import environ
 
 
-logger = logging.getLogger(__name__)
```
raphendyr: Keep the `logging.getLogger(__name__)`. It's better as then we can show the python module creating the log entry. Alternative (typically recommended) would be to write `logging.getLogger('apluslms_roman.utils.path_mapping')`, but as `__name__` returns that string, it's better. The recommendation presumes that names and modules might not match 100% and if the file is moved, then the logging name stays the same. I disagree with those arguments and thus prefer `__name__`.

If the file is executable (includes `if __name__ == '__main__'`, then the name for the logger needs to be string, so we don't get logger named `__main__`.

JohnDing1995: Ok, I restore `logging.getLogger( __name__)`  now

apluslms_roman/backends/docker.py
```diff
@@ -22,14 +25,28 @@ class DockerBackend(Backend):
...
+
+        # false values: 0, false, '', None
+        # true values: 1, true, "yes", unset
+        tls_verify = bool(env.get('tls_verify', False))
+        cert_path = env.get('cert_path') or None
+        if tls_verify or cert_path:
+            if not cert_path:
+                cert_path = os.path.join(os.path.expanduser('~'), '.docker')
+            params['tls'] = docker.tls.TLSConfig(
+                client_cert=(os.path.join(cert_path, 'cert.pem'), os.path.join(cert_path, 'key.pem')),
```
raphendyr: This file already has `from os.path import join`, so these could be just `join(cert_patj, 'cert.pem')` and so on.

apluslms_roman/backends/docker.py
```diff
@@ -22,14 +25,28 @@ class DockerBackend(Backend):
...
+        if 'timeout' in env:
+            params['timeout'] = env['timeout']
+
+        # false values: 0, false, '', None
+        # true values: 1, true, "yes", unset
+        tls_verify = bool(env.get('tls_verify', False))
+        cert_path = env.get('cert_path') or None
+        if tls_verify or cert_path:
+            if not cert_path:
+                cert_path = os.path.join(os.path.expanduser('~'), '.docker')
```
raphendyr: due to comment below, you need to add `expanduser` to import line `from os.path import join`.

apluslms_roman/backends/docker.py
```diff
@@ -22,14 +25,28 @@ class DockerBackend(Backend):
...
+            params['tls'] = docker.tls.TLSConfig(
+                client_cert=(os.path.join(cert_path, 'cert.pem'), os.path.join(cert_path, 'key.pem')),
+                ca_cert=os.path.join(cert_path, 'ca.pem'),
+                verify=tls_verify,
+                ssl_version=env.get('tls_ssl_version'),
+                assert_hostname=tls_verify and env.get('tls_assert_hostname'),
+            )
+        return DockerClient(**params)
```
raphendyr: You could have used `docker.DockerClient`. For some reason, it's how all docker stuff was done in this file, except `Mount`.

apluslms_roman/builder.py
```diff
@@ -1,14 +1,20 @@
...
-from os.path import isdir, join
+import logging
+from os import getuid, getegid
+from os.path import isdir
 
-from apluslms_yamlidator.utils.decorator import cached_property
 from apluslms_yamlidator.utils.collections import OrderedDict
+from apluslms_yamlidator.utils.decorator import cached_property
 
+from apluslms_roman.utils.path_mapping import load_from_env
```
raphendyr: Prefer `from .utils.path_mapping import ..`

apluslms_roman/builder.py
```diff
@@ -60,11 +65,13 @@ def __init__(self, backend_class=None, settings=None):
...
         name = getattr(backend_class, 'name', None) or backend_class.__name__.lower()
         env_prefix = name.upper() + '_'
-        env = {k: v for k, v in environ.items() if k.startswith(env_prefix)}
+        env = load_from_env(env_prefix, '.')
+        logger.debug("env without reading global config:%s", env)
         if settings:
             for k, v in settings.get(name, {}).items():
                 if v is not None and v != '':
-                    env[env_prefix + k.replace('-', '_').upper()] = v
+                    env[k] = v
```
raphendyr: This for loop can be replaced with:

```
if settings:
  env.update(settings.get(name, {}))
```

apluslms_roman/builder.py
```diff
@@ -78,4 +85,4 @@ def version_info(self):
         return self.backend.version_info()
 
     def create_builder(self, *args, **kwargs):
-        return Builder(self, *args, **kwargs)
+        return Builder(self, *args, **kwargs)
```
raphendyr: Where did the new line go?

apluslms_roman/schemas/roman_settings-v1.0.yaml
```diff
@@ -24,6 +24,10 @@ properties:
     type: object
     additionalProperties: false
     properties:
+      directory_map:
+        title: docker conatiner-host machine path mapping
```
raphendyr: typo in `conatiner`

apluslms_roman/utils/path_mapping.py
```diff
@@ -0,0 +1,70 @@
+from pathlib import PurePosixPath
+import json
+import logging
+import re
+from os import environ
```
raphendyr, line 4: Can you sort these.

apluslms_roman/utils/path_mapping.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+logger = logging.getLogger(__name__)
+json_re = re.compile(r'^(?:["[{]|(?:-?[1-9]\d*(?:\.\d+)?|null|true|false)$)')
+
+
+def get_host_path(original, mapping):
+    ret = original
+    orig_path = PurePosixPath(original)
+    for k, v in mapping.items():
```
raphendyr: Use longer variable names than just k and v. They are some times ok in comprehensions, but this loop isn't very clear about their meaning,

apluslms_roman/utils/path_mapping.py
```diff
@@ -0,0 +1,70 @@
...
+            relative_path = orig_path.relative_to(k)
+            ret = PurePosixPath(v).joinpath(relative_path)
+            return str(ret)
+        except ValueError:
+            logger.warning("Error when composing new path!")
+            pass
+    return str(ret)
+
+
+def get_pair_form_env(key, json_str, read_key=None):
```
raphendyr: This doesn't seem to be used by anything anymore?

apluslms_roman/utils/path_mapping.py
```diff
@@ -0,0 +1,70 @@
...
+        try:
+            ret = json.loads(json_str)
+            if isinstance(ret, dict):
+                return ret
+        except json.decoder.JSONDecodeError:
+            logger.warning("Error, check your json string")
+    return json_str
+
+
+def env_value_to_dict(json_str):
```
raphendyr: This seem to be only used in tests.

apluslms_roman/utils/path_mapping.py
```diff
@@ -0,0 +1,70 @@
...
+            root = key_list[0]
+            if root not in ret:
+                ret[root] = {}
+            ret[root][key_list[1]] = v
+        else:
+            ret[k] = v
+    return ret
+
+
+def load_from_env(env_prefix=None, sep=None, decode_json=True):
```
raphendyr: This on the other hand, doesn't have any tests.

tests/utils/test_path_mapping.py
```diff
@@ -0,0 +1,63 @@
...
+class TestJsonRegex(unittest.TestCase):
+
+    def test_loadable_match(self):
+        for i, case in enumerate(test_case_loadable):
+            with self.subTest(i=i):
+                self.assertTrue(json_re.match(case)is not None, msg="Testing:{}".format(case))
+
+    def test_not_loadable_not_match(self):
+        for i, case in enumerate(test_case_not_loadable):
+            with self.subTest(i=i):
```
raphendyr: You could use `with self.subTest(non_json=case):`

tests/utils/test_path_mapping.py
```diff
@@ -0,0 +1,63 @@
...
+                with self.assertRaises(ValueError, msg="Testing:{}".format(case)):
+                    loads(case)
+
+
+class TestJsonRegex(unittest.TestCase):
+
+    def test_loadable_match(self):
+        for i, case in enumerate(test_case_loadable):
+            with self.subTest(i=i):
```
raphendyr: You could use `with self.subTest(json=case):`

apluslms_roman/backends/docker.py
```diff
@@ -22,14 +25,28 @@ class DockerBackend(Backend):
...
+        params = {
+            'base_url': env.get('host'),
+            'version': env.get('version'),
+        }
+        if 'timeout' in env:
+            params['timeout'] = env['timeout']
+
+        # false values: 0, false, '', None
+        # true values: 1, true, "yes", unset
+        tls_verify = bool(env.get('tls_verify', False))
```
raphendyr: Based on the logic in the comments, we need to have `True` as default, instead of `False`.

apluslms_roman/backends/docker.py
```diff
@@ -22,14 +25,28 @@ class DockerBackend(Backend):
...
-        return docker.from_env(environment=env, **kwargs)
+        params = {
+            'base_url': env.get('host'),
+            'version': env.get('version'),
+        }
+        if 'timeout' in env:
+            params['timeout'] = env['timeout']
+
+        # false values: 0, false, '', None
+        # true values: 1, true, "yes", unset
```
raphendyr: Can you move the `unset` to false values in the comments. It actually belongs there.

apluslms_roman/builder.py
```diff
@@ -33,7 +33,7 @@ def get_steps(self, refs: list = None):
     def build(self, step_refs: list = None):
         backend = self._engine.backend
         observer = self._observer
-        steps = self.get_steps(step_refs) # NOTE: may raise KeyError or IndexError
+        steps = self.get_steps(step_refs)  # NOTE: may raise KeyError or IndexError
```
raphendyr: Drop this change.

apluslms_roman/builder.py
```diff
@@ -18,7 +19,6 @@ def __init__(self, engine, config, observer=None):
         self._engine = engine
         self._observer = observer or StreamObserver()
 
-
```
raphendyr: Drop this change.

apluslms_roman/backends/docker.py
```diff
@@ -41,9 +58,12 @@ def _run_opts(self, task, step):
...
+        path = self.remap_path(task.path)
+
+        logger.debug("Final path is:%s", path)
         # mounts and workdir
         if step.mnt:
-            opts['mounts'] = [Mount(step.mnt, task.path, type='bind', read_only=False)]
+            opts['mounts'] = [Mount(step.mnt, path, type='bind', read_only=False)]
             opts['working_dir'] = step.mnt
         else:
             wpath = self.WORK_PATH
```
raphendyr, line 69: This else is still using `task.path` instead of `path`

apluslms_roman/backends/__init__.py
```diff
@@ -112,3 +114,9 @@ def verify(self):
 
     def version_info(self):
         pass
+
+    def remap_path(self, path):
+        map_ = self.environment.environ.get('directory_map', {})
+        logger.debug("get mapping from environment:%s", map_)
+        map_ = dict(map_) if len(map_) == 0 else map_
```
raphendyr: What this is supposed to do?

apluslms_roman/backends/__init__.py
```diff
@@ -112,3 +114,9 @@ def verify(self):
 
     def version_info(self):
         pass
+
+    def remap_path(self, path):
```
raphendyr: this would be better named as `get_host_path`.

apluslms_roman/backends/__init__.py
```diff
@@ -1,19 +1,21 @@
...
 
 from ..observer import BuildObserver
-
+from ..utils.path_mapping import get_host_path
 
 BACKENDS = {
     'docker': 'apluslms_roman.backends.docker.DockerBackend',
 }
 
-
```
raphendyr: This is like a class definition, and thus should be surrounded by 2 new lines. Thus, don't remove this line.

apluslms_roman/backends/__init__.py
```diff
@@ -1,19 +1,21 @@
...
 BuildTask = namedtuple('BuildTask', [
     'path',
     'steps',
 ])
 
+logger = logging.getLogger(__name__)
```
raphendyr: This should be the first thing after imports. Thus, imports > logger > BACKENDS > BuildTask ..

apluslms_roman/utils/path_mapping.py
```diff
@@ -0,0 +1,48 @@
...
+import logging
+import re
+from os import environ
+from pathlib import PurePosixPath
+
+logger = logging.getLogger(__name__)
+json_re = re.compile(r'^(?:["[{]|(?:-?[1-9]\d*(?:\.\d+)?|null|true|false)$)')
+
+
+def get_host_path(original, mapping):
```
raphendyr: You can move this to the method in Backend. I don't see a need for it to be here as it's only:

```
if not mapping:
  return original
path = PurePosixPath(original)
for container, host in mapping.items():
    try:
        return str(PurePosixPath(host, path.relative_to(container)))
        # or (sadly, join understands pathlib objects only starting in py 3.6)
        # os.path.join(host, str(path.relative_to(container)))
    except ValueError:
        continue
raise ValueError("Unable to map path '%s' to any backend host path" % (path))
```

apluslms_roman/utils/path_mapping.py
```diff
@@ -0,0 +1,48 @@
...
+            relative_path = orig_path.relative_to(k)
+            ret = PurePosixPath(v).joinpath(relative_path)
+            return str(ret)
+        except ValueError:
+            logger.warning("Error when composing new path!")
+            pass
+    return str(ret)
+
+
+def nest_dict(flat_dict, sep):
```
raphendyr: This could be implemented as (and moved into load_from_env):

```
from apluslms_yamlidator.document import find_ml

nested = {}
for keys, value in flat_dict.items():
    dict_, key = find_ml(nested, keys, create_dicts=True)
    dict_[key] = value
return nested
```

Sadly, that doesn't support `sep`, but I don't see there would bee ever need to support other than `.`.

ref: https://github.com/apluslms/roman/blob/milestone0.3/apluslms-yamlidator/apluslms_yamlidator/document.py#L30

raphendyr: I might end up extending the find_ml to support key strings like `foo[].bar[]` some day :)

apluslms_roman/backends/__init__.py
JohnDing1995: Should be removed

apluslms_roman/backends/__init__.py
JohnDing1995: I'll switch it's function name with the function `git_host_path()` in `utils.py`

apluslms_roman/builder.py
JohnDing1995: In PEP: 
Method definitions inside a class are surrounded by a single blank line.
https://www.python.org/dev/peps/pep-0008/#blank-lines
Shouldn't we make it only one blank line?

raphendyr: My bad. I think I looked this was in root. Yeah, it's a correct change and I didn't notice this in earlier PR (when it was introduced).

raphendyr:
After a long though, I have come to conclusion. I think we will move forward with creating a volume, then uploading the course material to the volume and running the builds. To speed up this in iterative builds, we can use librsync or similar to upload only changed files.

Thus, before closing this PR, we need to:
 * Split the docker backend features (better docker object creation)
 * Use the json_re etc. when reading environment variables for roman configuration (the generic case)

In addition, we will look into implementing the volume stuff. Until that, the current system works for local building (the currently supported usecase).

PR #9, opened 2019-06-04
commands for roman client

ppessi:


apluslms_roman/configuration.py
```diff
@@ -56,16 +62,33 @@ def find_from(cls, path, name=None, prefixes=None):
                     break
         else:
             if is_legacy_configuration(path):
-                container = cls.Container(config, allow_missing=True)
+                container = cls.Container(join(path, 'roman.yml'), allow_missing=True)
```
raphendyr: Maybe we could use `DEFAULT_FILENAME` here and define that in the globals.

apluslms_roman/configuration.py
```diff
@@ -1,19 +1,27 @@
...
 from apluslms_yamlidator.document import Document
 from apluslms_yamlidator.utils.collections import OrderedDict
 from apluslms_yamlidator.utils.version import Version
 
 from .utils.translation import _
 
-DEFAULT_NAME = 'course'
+DEFAULT_NAME = 'roman'
+DEFAULT_NAMES = ('roman', 'course')
 DEFAULT_PREFIXES = ('yml', 'yaml', 'json')
```
raphendyr: We could define `DEFAULT_FILENAME = "%s.%s" % (DEFAULT_NAMES[0], DEFAULT_PREFIXES[0])` here.

apluslms_roman/configuration.py
```diff
@@ -1,19 +1,27 @@
...
+from os.path import basename, join, isdir, isfile, relpath
 
 from apluslms_yamlidator.document import Document
 from apluslms_yamlidator.utils.collections import OrderedDict
 from apluslms_yamlidator.utils.version import Version
 
 from .utils.translation import _
 
-DEFAULT_NAME = 'course'
+DEFAULT_NAME = 'roman'
```
raphendyr: This is not used within the file.

apluslms_roman/configuration.py
```diff
@@ -1,19 +1,27 @@
+from itertools import chain
 from os import listdir
-from os.path import join, isdir, isfile
+from os.path import basename, join, isdir, isfile, relpath
```
raphendyr: `relpath` is not used within the file

apluslms-yamlidator/apluslms_yamlidator/document.py
```diff
@@ -192,7 +193,9 @@ def load(cls, path, version=None, **kwargs):
         try:
             document = container.get_latest(max_version=version)
         except KeyError:
-            document = cls(container, None, Dict(), version)
+            document = cls(container, None, Dict(version=str(cls.version)), version)
```
raphendyr: I presume there should be `str(version)` instead of `str(cls.version)`.

apluslms_roman/cli.py
```diff
@@ -379,6 +413,142 @@ def config_print_action(context):
...
+    try:
+        config_path = RomanConfig.find_from(cwd).path
+        # if the folder is legacy, find_from won't error
+        # so we need to check if the file actually exists
+        if not isfile(config_path):
+            raise RomanConfigError
+        exit(1, "{} already exists in {}.\nNo new Roman config created.".format(
+            basename(config_path), cwd)
+        )
+    except RomanConfigError:
```
raphendyr: I would prefer this logic to be in `configuration.py`, so it would probably be easier to use from the GUI too.

That said, I think this should be easy as `RomanConfig.load(join(getcwd(), RomanConfig.DEFAULT_FILENAME, allow_missing=True).save()`. We don't need to hangle legacy_config in this case. The whole code should be something like:

```
try:
  try:
    RomanConfig.find_from(getcwd()).save(create=True) # create=True uses 'x' instead of 'w' for `open`
  except RomanConfigError:
    RomanConfig.create(join(getcwd(), RomanConfig.DEFAULT_FILENAME))
except FileExistsError:
  exit(1, "file exists")
```

And create classmethod in Document:

```
@classmethod
def create(self, filename):
  document = cls.load(filename, allow_missing=True)
  document.save(create=True) # create=True uses 'x' instead of 'w' for `open`
  return document
```

`find_from` should set a real path when it loads legacy configuration in memory, so we can use `save`. In addition, `DEFAULT_PREFIXES`, `DEFAULT_NAMES` and `DEFAULT_FILENAME` should probably be defined in the RomanConfig class, so they are easier to use from outside. And that `create=True` needs to be implemented.

apluslms_roman/configuration.py
```diff
@@ -22,30 +30,28 @@
...
+class RomanConfigError(Exception):
     pass
 
 def is_legacy_configuration(path):
     """Checks if path is legacy a-plus-rst-tools course"""
     return isfile(join(path, 'conf.py')) and isfile(join(path, 'Makefile'))
 
-class CourseConfig(Document):
-    name = 'course_config'
+class RomanConfig(Document):
```
raphendyr: Actially, the class and the error class probably should ne `ProjectConfig` and `ProjectConfigError`. Furthermore, `RomanSettings` probably should be `GlobalSettings`, thus making these two classes clearly distinct.

apluslms-yamlidator/apluslms_yamlidator/document.py
raphendyr: This should also handle the case when version is None. In addition, it should use `version_key`. This probably should also support creating empty documents with initialization function, as the schema might require some values (e.g. steps in the project config).

So maybe something:

```
except KeyError:
  if version is not None:
    document = cls(container, None, Dict(), version)
    documnet.initialize_data()
    document.validate() # should validate as long as initialize_data ensures the initial date is there
 ```

and in Document:

```
def initialize_data(self):
  vkey = self.version_key
  if vkey:
    self._data[vkey] = str(self.version)
```

File: apluslms_roman/backends/docker.py [hunk 8]
```diff
@@ -75,6 +77,8 @@ def build(self, task, observer):
             observer.start_step(step)
             opts = self._run_opts(task, step)
             observer.manager_msg(step, "Running container {}:".format(opts['image']))
+            if isinstance(opts["command"], ChangesList):
```
raphendyr: This is not a good place to handle this. I think the best place is in `BuildStep.__init__`: https://github.com/apluslms/roman/blob/f2ec26f6e41091d51d7304ded3706adfca231c28/apluslms_roman/backends/__init__.py#L51

Both env and cmd should be handled there. For example:

```
self.img = clean_image_name(img)
self.cmd = cmd if cmd is None or isinstance(cmd, str) else list(cmd)
self.env = {} if env is None else dict(env)
```

and remove all those actions from `from_config`.

apluslms_roman/cli.py
```diff
@@ -379,6 +413,142 @@ def config_print_action(context):
         print("---\n# project config")
         yaml_dump(get_config(context)._data, stdout)
 
+def init_config(context):
```
raphendyr: The function should be named `init_action` to match other cli actions.

apluslms_roman/cli.py
```diff
@@ -379,6 +413,142 @@ def config_print_action(context):
...
+        )
+    except RomanConfigError:
+        filename = DEFAULT_NAME + ".yml"
+        config = LEGACY_CONFIG if is_legacy_configuration(cwd) else DEFAULT_CONFIG
+        filepath = path_join(cwd, filename)
+        with open(filepath, 'w') as outfile:
+            yaml_dump(config, outfile)
+        print("{} successfully created in {}".format(filename, cwd))
+
+def set_args(context):
```
raphendyr: `config_set_action`

apluslms_roman/cli.py
```diff
@@ -379,6 +413,142 @@ def config_print_action(context):
...
+        raise Exception("Choose either user settings or project settings")
+    if context.args.user:
+        document = context.settings
+    else:
+        document = context.config
+
+    schema = document.validator.schema
+    for val in context.args.values:
+        try:
+            key, val = val.split('=')
```
raphendyr: Using `val.split('=', 1)` would allow `FOOBAR=FOO=FOO` and resuls `key="FOOBAR"; val="FOO=FOO"`.

raphendyr: Document or RomanConfig could have interface to set a key, value pair with type casting or to set list of key, value pairs with type casting. As side effect, this complex code would not reside in `cli.py`.

apluslms_roman/cli.py
```diff
@@ -379,6 +413,142 @@ def config_print_action(context):
...
+                elif val_type == "boolean":
+                    if val.lower == "true" or val == "1":
+                        val = True
+                    elif val.lower == "false" or val == "0":
+                        val = False
+                    else:
+                        raise TypeError("'" + key + "' should be a boolean.")
+            except KeyError:
+                raise KeyError("'" + key + "' is not a key found in the schema.")
+            document.mlset(key, val)
```
raphendyr: E.g. `document.mlset_cast(key, val)`

apluslms_roman/cli.py
```diff
@@ -379,6 +413,142 @@ def config_print_action(context):
...
+        "mnt": args.mnt,
+        "env": env,
+        "name": args.name
+    }
+    nulls = [k for k in step if not step[k]]
+    for k in nulls:
+        del step[k]
+    for k in step:
+        if k != "env":
+            step[k] = step[k][0]
```
raphendyr: Is this here because there is `nargs=1` instead of `nargs=None`?

apluslms_roman/cli.py
```diff
@@ -379,6 +413,142 @@ def config_print_action(context):
...
+        except ValueError:
+            raise ValueError("env is a dict, so values need to be in key=val format, e.g. a=1 b=2")
+    step = {
+        "img": args.img,
+        "cmd": args.cmd,
+        "mnt": args.mnt,
+        "env": env,
+        "name": args.name
+    }
+    nulls = [k for k in step if not step[k]]
```
raphendyr: `step = {k: v for k, v in step.items() if v}` is probably simpler than removing empties.

apluslms_roman/cli.py
```diff
@@ -379,6 +413,142 @@ def config_print_action(context):
...
+        del step[k]
+    for k in step:
+        if k != "env":
+            step[k] = step[k][0]
+    config = get_config(context)
+    if "name" in step:
+        other_names = [s["name"] for s in [s.get_data() for s in config.steps] if "name" in s]
+        if step["name"] in other_names:
+            raise ValueError("A step with the name '" + step["name"] + "' already exists.")
+    config.steps.append(step)
```
raphendyr: RomanConfig probably should have a steps add/del interface, which would take care of duplicate names for example.

apluslms_roman/cli.py
```diff
@@ -379,6 +413,142 @@ def config_print_action(context):
...
+            if "name" in s.get_data() and s.get_data()["name"] == name
+        ][0]
+        if not confirm_del(step):
+            return
+        config.steps.remove(step)
+    config.validate()
+    config.save()
+    print("Step successfully removed from config.")
+
+def confirm_del(step):
```
raphendyr: this could be inside `del_step` and it should be possible to skip it with `-f` / `--force` flag.

raphendyr:
Updated milestone0.3 so it doesn't build the GUI component. That should "fix" travis ci tests.

apluslms_roman/configuration.py
ppessi: it was used in cli but I suppose it isn't needed if we have default_filename

apluslms-yamlidator/apluslms_yamlidator/document.py
ppessi: what should be done if version_key is empty?

ppessi: what should be in initialize? all data required for passing validation is version

apluslms_roman/configuration.py
ppessi: yes, that would make it a lot clearer

apluslms_roman/backends/docker.py
ppessi: I'm assuming this also applies to everything in the milestone0.3_global_vars branch?

apluslms_roman/cli.py
ppessi: I can't remember anything about this :') well, shouldn't be too hard to find out...

ppessi: okay yeah I must've assumed nargs=1 would mean that there is one argument. which in my opinion is a completely reasonable assumption...?

apluslms-yamlidator/apluslms_yamlidator/document.py
raphendyr: If version key is empty, then the schema probably doesn't require it. In that case, it could even violate the schema.

For RomanConfig/ProjectConfig we could add the default or the legacy configuration. 50/50 if that is a good idea, though. And there is no need for subclass to overwrite the mehtod if they don't need to.

If the schema is written well, then the defaults should ensure the document is always valid.

apluslms_roman/backends/docker.py
raphendyr: I presume so.

apluslms_roman/cli.py
ppessi: I personally would find `overwrite=False` more intuitive, would that work?

apluslms-yamlidator/apluslms_yamlidator/document.py
ppessi: so... add defaults to version in the schemas?

apluslms_roman/cli.py
raphendyr: True, that seems better. I maybe prefer default values to be False or None, but I think this is good place to deviate from that. `create=True` is misleading at least. It that would be used, then it should be `fail_if_exists` or `require_create` and both are a bit unclear, thus `overwrite=True` or `allow_overwrite=True` is probably the best.

apluslms-yamlidator/apluslms_yamlidator/document.py
raphendyr: Version is a bit special child and it's probably better to handle this way and bit specially. It could be even arguable to drop version_key from data before validation and drop it from schemas. Would make those a bit more clean.

apluslms-yamlidator/apluslms_yamlidator/document.py
```diff
@@ -322,10 +355,49 @@ def mlset(self, keys, value):
...
+        schema_keys = list(chain.from_iterable(zip(['properties'] * len(key_list), key_list)))
+
+        container, key = find_ml(self.validator.schema, schema_keys)
+
+        if key in container:
+            container = container[key]
+            if "type" in container:
+                val_type = container["type"]
+                if val_type == "integer":
+                    if value.isdigit():
```
raphendyr: Hmm. Noticed that `-123` doesn't pass this. I think we can just call `int(value)` and if ValueError is raised, we just propagate that up.

If type casting is wrapped to try-except, we could add the expected value type to the exception:

```
try:
  value = int(value)
except ValueError as e:
  e.value_type = val_type
  raise e
```

This way we can move the UI error message to cli.py file.

apluslms-yamlidator/apluslms_yamlidator/document.py
```diff
@@ -322,10 +355,49 @@ def mlset(self, keys, value):
...
+                        value = int(value)
+                    else:
+                        raise TypeError(type_msg.format(keys, val_type))
+                elif val_type == "boolean":
+                    if value.lower == "true" or value == "1":
```
raphendyr: Converting a string to a boolean, could be done in utility file. ValueError is better Exception and probably more suited for that.

For comparison, the similar logic in configparser:
https://github.com/python/cpython/blob/master/Lib/configparser.py#L826
https://github.com/python/cpython/blob/master/Lib/configparser.py#L1159
https://github.com/python/cpython/blob/master/Lib/configparser.py#L598

For now, 0, 1, true and false is good. In a long run, supporting language specific values would be cool.

apluslms-yamlidator/apluslms_yamlidator/document.py
```diff
@@ -38,6 +40,16 @@ def find_ml(current, keys):
             current = current.setdefault(key, {})
     return current, keys[-1]
 
+def update(data, new_data):
+    for key in new_data:
+        if key in data:
+            if isinstance(data[key], list) and isinstance(new_data[key], list):
```
raphendyr: More complete list of type testing is `if instance(plaah, (list, tuple, Sequence)) and not isinstance(plaah, (str, bytes))`

apluslms-yamlidator/apluslms_yamlidator/document.py
```diff
@@ -38,6 +40,16 @@ def find_ml(current, keys):
...
+                data[key].extend(new_data[key])
+                continue
+            if isinstance(data[key], dict) and isinstance(new_data[key], dict):
```
raphendyr: Should test that `isinstance(plaah, (dict, MutableMapping)`, new_data of course could be just `Mapping`.

apluslms-yamlidator/apluslms_yamlidator/document.py
```diff
@@ -38,6 +40,16 @@ def find_ml(current, keys):
             current = current.setdefault(key, {})
     return current, keys[-1]
 
+def update(data, new_data):
```
raphendyr: This function could be implemented in `.utils.collections`.

apluslms-yamlidator/apluslms_yamlidator/document.py
```diff
@@ -322,10 +355,49 @@ def mlset(self, keys, value):
...
     def mlsetdefault(self, keys, value):
         container, key = find_ml(self._data, keys)
         return container.setdefault(key, value)
+
+    def mldel(self, keys):
+        container, key = find_ml(self._data, keys)
+        del container[key]
+
+    def recursive_update(self, new_data):
+        update(self._data, new_data)
```
raphendyr: If you say `update(self, new_data)` then both _data and _work are updated and dirty bit is set correctly.

apluslms_roman/backends/__init__.py
```diff
@@ -38,17 +39,21 @@ def from_config(cls, index, data):
...
                 data.get('name'),
             )
         else:
             return cls(index, clean_image_name(data))
 
     def __init__(self, ref, img, cmd=None, mnt=None, env=None, name=None):
-        self.img, self.cmd, self.mnt, self.env, self.name, self.ref = img, cmd, mnt, env, name, ref
+        self.img = clean_image_name(img)
+        self.env = dict(env) if env else {}
+        if isinstance(cmd, ChangesList):
```
raphendyr: `if not isinstance(cmd, str)`. Try to avoid importing ChangesList to different locations. In addition, it might be better to cast cmd to `tuple`, so it isn't so easy to edit it by accident.

apluslms_roman/backends/__init__.py
```diff
@@ -38,17 +39,21 @@ def from_config(cls, index, data):
...
+            cmd = list(cmd)
+        self.cmd, self.mnt, self.name, self.ref = cmd, mnt, name, ref
```
raphendyr: It's probably a good time to split all of these to their own lines.

p.s. cmd could be written as `self.cmd = cmd if isinstance(cmd, str) else tuple(cmd)`.

apluslms_roman/backends/docker.py
```diff
@@ -1,5 +1,6 @@
 import docker
 from os.path import join
+from sys import exit
```
raphendyr: ??

apluslms_roman/cli.py
```diff
@@ -14,8 +14,9 @@
 
 from . import __version__
 from .builder import Engine
-from .configuration import CourseConfigError, CourseConfig
-from .settings import RomanSettings
+from .configuration import (ProjectConfigError, ProjectConfig, LEGACY_CONFIG, DEFAULT_CONFIG,
```
raphendyr: Write all imported keys on their own lines and sort them (though, globals in start(. So:
```
from .configuration import (
    DEFAULT_CONFIG,
    LEGACY_CONFIG,
    ProjectConfig,
    ProjectConfigError,
    is_legacy_configuration, # don't forget this comma
)
´``

File: apluslms_roman/cli.py [hunk 27]
```diff
@@ -197,17 +198,19 @@ def parse_actioncontext(parser):
...
 
     # load settings
-    allow_missing = args.config is None
-    config = args.config if args.config is not None else RomanSettings.get_config_path()
-
+    project = "project" in args and args.project
+    allow_missing = args.config is None and not project
+    config = (
+        args.config if args.config is not None
+        else getcwd() + "/" + GlobalSettings.name + ".yml" if project
```
raphendyr: I'm not sure what this is supposed to do here? Global settings file should be selected with `--config` flag.

apluslms_roman/configuration.py
```diff
@@ -56,16 +65,67 @@ def find_from(cls, path, name=None, prefixes=None):
...
+        document = cls.load(filename, allow_missing=True)
+        document.save(overwrite=False)
+        return document
+
+    def initialize_data(self):
+        self.recursive_update(
+            LEGACY_CONFIG if is_legacy_configuration(dirname(self.path))
+            else DEFAULT_CONFIG
+        )
+        self._data[self._version_key] = str(self.version)
```
raphendyr: `super().initialize_data()` ;)

apluslms_roman/configuration.py
```diff
@@ -56,16 +65,67 @@ def find_from(cls, path, name=None, prefixes=None):
...
+            return cls.load(config)
+        if '.' not in basename(config):
+            for prefix in cls.DEFAULT_PREFIXES:
+                filename = '%s.%s' % (config, prefix)
+                if isfile(filename):
+                    return cls.load(filename)
+        raise ProjectConfigError("Given file '{}' doesn't exist.".format(config))
+
+    @classmethod
+    def create(cls, filename):
```
raphendyr: Can you move this to Document class.

apluslms_roman/configuration.py
```diff
@@ -56,16 +65,67 @@ def find_from(cls, path, name=None, prefixes=None):
...
+        )
+        self._data[self._version_key] = str(self.version)
+
+    def add_step(self, step):
+        if "name" in step:
+            other_names = [s["name"] for s in [s.get_data() for s in self.steps] if "name" in s]
+            if step["name"] in other_names:
+                raise ValueError("A step with the name '" + step["name"] + "' already exists.")
+        self.steps.append(step)
+        self.validate()
```
raphendyr: Validate and save should ne on cli.py side

apluslms-yamlidator/apluslms_yamlidator/document.py
ppessi: in which languages are there some other options than those...?

apluslms_roman/settings.py
```diff
@@ -112,10 +112,11 @@ def update_from_namespace(self, namespace, store=False):
                     set_(option, value)
 
 
-class RomanSettings(ArgumentSettings):
+class GlobalSettings(ArgumentSettings):
     name = 'roman_settings'
     schema = name
     version = Version(1, 0)
+    _version_key = "version"
```
raphendyr: This fixes something, so what was the problem? I have a feeling that there is a bug in the document metaclass, which would solve it..

apluslms_roman/configuration.py
```diff
@@ -56,16 +65,67 @@ def find_from(cls, path, name=None, prefixes=None):
...
+        self.save()
+
+    def del_step(self, step_ref):
+        step = self.get_step(step_ref)
+        self.steps.remove(step)
+
+    def get_step(self, step_ref):
+        if step_ref.isdigit():
+            return self.steps[int(step_ref)]
+        names = [s["name"].lower() for s in self.steps if "name" in s]
```
raphendyr: If this would be `return {s['name']: s for s ub self.steps if 'name' in s}[step_ref]`, then we would get two benefits. First, we would just iterate steps only once. Second, we would raise KeyError, when the step is not found. Then in cli we can except IndexError and KeyError separately and have different messages for them.

The dictionary could be created in a property, making it easy to use in other methods too.

apluslms_roman/configuration.py
```diff
@@ -56,16 +65,67 @@ def find_from(cls, path, name=None, prefixes=None):
...
+        if step_ref not in names:
+            raise ValueError("There is no step named '{}'".format(step_ref))
+        return [s for s in self.steps if "name" in s and s["name"].lower() == step_ref][0]
```
raphendyr: FYI: `next(s for s in self.steps if "name" in s and s["name"].lower() == step_ref)` would be a faster code.

apluslms_roman/configuration.py
raphendyr: p.s. with single and double quotes we can indicate meta info. Strings in single quotes are keys and similar code related values, while double quoted strings are messages and other strings typically shown to the user. Typically, single quotes don't include spaces and if the string contains space, then it's double quoted most of the time.

apluslms-yamlidator/apluslms_yamlidator/document.py
raphendyr: For Finnish "kyllä" and "ei". As this is parsing user input from command line, it could be argued that when cli is used with `LC_CTYPE=fi_FI` we should parse also those strings. This can bee looked when (and if) cli is translated.

apluslms-yamlidator/apluslms_yamlidator/document.py
ppessi: what's the point of `e.value_type = val_type`? doesn't change the error message at all.

apluslms-yamlidator/apluslms_yamlidator/document.py
ppessi: ah as in real languages, not programming languages :p

apluslms-yamlidator/apluslms_yamlidator/document.py
ppessi: an unedited ValueError isn't very useful for the user either, it doesn't really tell where the problem is

apluslms-yamlidator/apluslms_yamlidator/document.py
ppessi: in which utility file should this be? if it's added to `__init__` it could be imported like `from utils import X`, right? module_resources kind of seems like a 'misc' file.

apluslms-yamlidator/apluslms_yamlidator/document.py
ppessi: we can probably assume that anything already in the data is a list?

ppessi: ah the dict thing was in the next comment

apluslms_roman/backends/docker.py
ppessi: leftover test stuff...

apluslms_roman/cli.py
ppessi: some old remnants that I didn't properly edit I guess...?

apluslms-yamlidator/apluslms_yamlidator/document.py
ppessi: ...missed the last sentence. I'm very good at reading these comments :')

apluslms-yamlidator/apluslms_yamlidator/document.py
raphendyr: `module_resources` contains stuff handling resources files bundled within python packages. It extens functionality provided by importlib.resources (or backported library).

You can put it in utils/__init__.py for now.

apluslms-yamlidator/apluslms_yamlidator/document.py
raphendyr: For example, ChangesList doesn't inherit from a list, but it does from Sequence. Sadly, so does str.

apluslms_roman/settings.py
ppessi: I mentioned this to you earlier, some stuff in the Document class requires version_key, but a Document can be created without a version_key

apluslms_roman/configuration.py
ppessi: I knew there should probably be a smarter way to do this but couldn't think of it :')
Yeah I know the general practices with quotes, I'm just so used to using double quotes I end up putting double quotes in a lot of places accidentally...

apluslms-yamlidator/apluslms_yamlidator/document.py
ppessi: okay. but no tuples in the original data?

apluslms-yamlidator/apluslms_yamlidator/document.py
ppessi: so `isinstance(new_data[key], (dict, MutableMapping, Mapping))`, or what exactly did you mean?

raphendyr: MutableMapping is always a subclass of Mapping. One of them is fine. `new_data` could be just a Mapping as we don't need to write to it, though I don't know any existing instances of Mapping, only MutableMappings. `data` needs to be mutable for us to be able to write to it.

apluslms-yamlidator/apluslms_yamlidator/document.py
raphendyr: Most likely no. Techincally tuples would be fine for `new_data`, but not for `data` (can't append to tuples)

apluslms_roman/settings.py
raphendyr: I started to update `DocumentMeta` to fix this.

raphendyr: I'm pretty sure this is now fixed in milestone0.3 in commit 30be357fc7e79ec50180f46b13295505ba79669b

apluslms_roman/configuration.py
raphendyr: This line should not be necessary as the `super().initialize_data()` should set `self[self.version_key]`, if it there is version_key.

apluslms-yamlidator/apluslms_yamlidator/document.py
```diff
@@ -193,10 +195,23 @@ def load(cls, path, version=None, **kwargs):
...
+    @classmethod
+    def create(cls, filename):
+        document = cls.load(filename, allow_missing=True)
+        document.save(overwrite=False)
+        return document
+
+    def initialize_data(self):
+        vkey = self.version_key
+        if vkey:
+            self._data[vkey] = str(self.version)
```
raphendyr: Hmm.. This probably should be `self[vkey] = str(self.version)`, so that the value in `self._work` is also set.

apluslms_roman/backends/__init__.py
```diff
@@ -38,17 +38,22 @@ def from_config(cls, index, data):
...
             return cls(index, clean_image_name(data))
 
     def __init__(self, ref, img, cmd=None, mnt=None, env=None, name=None):
-        self.img, self.cmd, self.mnt, self.env, self.name, self.ref = img, cmd, mnt, env, name, ref
+        self.img = clean_image_name(img)
+        self.env = dict(env) if env else {}
+        self.cmd = cmd if isinstance(cmd, str) else tuple(cmd)
+        self.mnt = mnt
+        self.name = name
+        self.ref = ref
```
raphendyr: I do prefer sorting these in the same order as function arguments, when that doesn't make code hard to read.

tests/document_test.py
```diff
@@ -0,0 +1,26 @@
+import unittest
```
raphendyr: As this is yamlidator test, this probably should be under `apluslms-yamlidator/tests`. I presume we would like to name it `apluslms-yamlidator/tests/utils/test_collections.py`, so that it matches the module structure inside the package and starts with `test`, which is the default pattern used by unittest library.

I added some tests as reference in 5b6098320cd45ba0dfe5f32fc3a1e0c6444100e1

apluslms-yamlidator/apluslms_yamlidator/document.py
ppessi: what is self._work? haven't used it for anything so far

apluslms-yamlidator/apluslms_yamlidator/document.py
```diff
@@ -38,7 +42,6 @@ def find_ml(current, keys):
             current = current.setdefault(key, {})
     return current, keys[-1]
 
-
```
raphendyr: I prefer to use two blank lines to separate root level functions etc., but often three between different classes.

> Surround top-level function and class definitions with two blank lines.
https://www.python.org/dev/peps/pep-0008/#blank-lines

apluslms-yamlidator/apluslms_yamlidator/document.py
```diff
@@ -280,6 +319,22 @@ def __repr__(self):
             self._data,
         )
 
+    def find_type(self, keys):
+        if isinstance(keys, str):
+            key_list = keys.split(".")
+        elif not isinstance(keys, list):
+            key_list = list(keys)
+
+        schema_keys = list(chain.from_iterable(zip(['properties'] * len(key_list), key_list)))
```
raphendyr: Possible alternative you can consider `list(chain.from_iterable(zip_longest((), key_list, fillvalue='properties')))`

apluslms-yamlidator/apluslms_yamlidator/document.py
```diff
@@ -322,10 +377,46 @@ def mlset(self, keys, value):
...
+            elif val_type == "boolean":
+                try:
+                    value = to_bool(value)
+                except ValueError as err:
+                    err.value_type = val_type
+                    raise err
+            elif val_type == "string":
+                try:
+                    value = int(value)
+                    print("mui")
```
raphendyr: I presume this is some testing code :)

apluslms-yamlidator/apluslms_yamlidator/document.py
```diff
@@ -322,10 +377,46 @@ def mlset(self, keys, value):
...
+            if val_type == "integer":
+                try:
+                    value = int(value)
+                except ValueError as err:
+                    err.value_type = val_type
+                    raise err
+            elif val_type == "boolean":
+                try:
+                    value = to_bool(value)
+                except ValueError as err:
```
raphendyr, line 417: If you move this between `if val_type:` and `if val_type == "integer":`, then you would only need one try..except. Both are fine though.

apluslms-yamlidator/apluslms_yamlidator/document.py
raphendyr: Changes dict contains three data storages: data, work and defaults. Work is merge between defaults and data. In addition, work is the dictionary which effects output from `__iter__` and so on. The work dict is what is seen when using a ChangesDict and the data is what goes to disk. `ChangesDict.__setitem__()` write value to both data and work, so the data is stored on disk and is also the value used. In contrast, `ChangesDict.setdefault()` updates defaults and also work only if the key doesn't exists there.

apluslms_roman/cli.py
raphendyr: nargs=1 and nargs=None both mean "there is only one argument", but the value returned is in different format. IF I read the code correctly, nargs=None produces a single value and nargs=1 produces a list (else branch in following code).

ref: https://github.com/python/cpython/blob/master/Lib/argparse.py#L2375

apluslms-yamlidator/apluslms_yamlidator/utils/__init__.py
```diff
@@ -0,0 +1,12 @@
+BOOLEANS = {
+    0: False, 1: True,
+    '0': False, '1': True,
+    'false': False, 'true': True,
+    'False': False, 'True': True
+}
+
+def convert_to_boolean(val):
+    if val not in BOOLEANS:
+        raise ValueError
```
raphendyr: A bit more generic version would be `raise ValueError("invalid literal for convert_to_boolean(): %r" % (val,))`. Would it be nice?

apluslms-yamlidator/apluslms_yamlidator/utils/__init__.py
```diff
@@ -0,0 +1,12 @@
+BOOLEANS = {
+    0: False, 1: True,
+    '0': False, '1': True,
+    'false': False, 'true': True,
+    'False': False, 'True': True
+}
+
+def convert_to_boolean(val):
+    if val not in BOOLEANS:
```
raphendyr, line 11: Technically this makes two queries to BOLEANS (relatively fast). I'm not sure if this alternative would be slower or faster..

```
try:
  return BOOLEANS[val]
except KeyError:
  raise ValueError(...)
```

apluslms-yamlidator/tests/utils/test_collections.py
```diff
@@ -0,0 +1,64 @@
...
+    def test_updates_dicts_recursive(self):
+        old = {'dict1': {'dict2': {'foo': 'bar'}}}
+        new = {'dict1': {'dict2': {'a': 1}, 'dict3': {'b': 2}}}
+        update(old, new)
+        self.assertEqual(sorted(list(old['dict1'].keys())), sorted(['dict2', 'dict3']))
+        self.assertEqual(old['dict1']['dict2'], {'foo': 'bar', 'a': 1})
+
+
+if __name__ == '__main__':
+    unittest.main()
```
raphendyr: I think this is not needed in test files, as the file can easily be run using `python3 -m unittest tests.utils.test_collections` or  `./setup.py test -s tests.utils.test_collections`

apluslms_roman/configuration.py
```diff
@@ -56,16 +65,61 @@ def find_from(cls, path, name=None, prefixes=None):
...
+        if step_ref.isdigit():
+            return self.steps[int(step_ref)]
+        return self.steps_by_name[step_ref]
+
     @property
     def dir(self):
         return self._container._dir
 
+    @property
+    def path(self):
```
raphendyr: Hmm... `Document` has same `.path`, so this should not be needed as it's duplicate, or am I missing someting?

apluslms_roman/configuration.py
```diff
@@ -56,16 +65,61 @@ def find_from(cls, path, name=None, prefixes=None):
...
+            for prefix in cls.DEFAULT_PREFIXES:
+                filename = '%s.%s' % (config, prefix)
+                if isfile(filename):
+                    return cls.load(filename)
+        raise ProjectConfigError("Given file '{}' doesn't exist.".format(config))
+
+    def initialize_data(self):
+        super().initialize_data()
+        self.recursive_update(
+            LEGACY_CONFIG if is_legacy_configuration(dirname(self.path))
```
raphendyr: Could `dirname(self.path)` be replaced with `self.dir`?

apluslms-yamlidator/apluslms_yamlidator/utils/__init__.py
ppessi: The time you gain from only calling val once isn't as big as the time you lose with the try except. Then again we can probably assume that most of the time an error won't be raised. But the method isn't going to be called that much (usually 0 or 1 times) so the difference isn't significant either way :p

raphendyr: Yep. Thus I think the current solutions is totally fine. Wanted to leave this for the record and as something I would like to benchmark some day. Reference implementation in configparser does even call `key.lower()` twice..

apluslms_roman/configuration.py
ppessi: ...yet another in the series of "stuff I did for some reason but who knows what that reason was"

apluslms-yamlidator/apluslms_yamlidator/document.py
```diff
@@ -2,13 +2,17 @@
 from abc import ABCMeta
 from copy import deepcopy
 from hashlib import sha1
+from itertools import chain, zip_longest
```
raphendyr, line 6: `zip_longest` is imported, but not used.  Was is supposed to be used?

apluslms-yamlidator/apluslms_yamlidator/document.py
```diff
@@ -325,10 +381,46 @@ def mlset(self, keys, value):
...
+            if val_type == "boolean":
+                try:
+                    value = to_bool(value)
+                except ValueError as err:
+                    err.value_type = val_type
+                    raise err
+            elif val_type == "integer" or val_type == "string":
+                try:
+                    value = int(value)
+                    # if conversion is successfull, the value is not a string
```
raphendyr: I'm not sure if this is a good idea. There might be cases where one wants to add integer to a string field. For example, when adding environment variables. I would presume that this is less than 1% of the case, but still.

So, which kind of cases this check should solve or what is the goal for it?

apluslms-yamlidator/apluslms_yamlidator/utils/collections.py
```diff
@@ -16,6 +16,19 @@
 else:
     from collections import OrderedDict
 
+def update(data, new_data):
```
raphendyr: I think this could be named `recursive_update` too.

apluslms_roman/backends/__init__.py
```diff
@@ -1,6 +1,7 @@
 from collections import namedtuple
 from collections.abc import Mapping
 
+from apluslms_yamlidator.utils.collections import ChangesList
```
raphendyr: This is not used in the file anymore.

apluslms_roman/cli.py
```diff
@@ -346,9 +383,10 @@ def build_action(context):
         steps = builder.get_steps()
         num_len = max(2, len(str(len(steps)-1)))
         name_len = max(4, len(max(step.name or "" for step in steps)))
+        name_len = max(4, max(len(s.name or "") for s in steps))
```
raphendyr: Replaces above line, but that old line is not removed.

apluslms-yamlidator/apluslms_yamlidator/document.py
ppessi: it was at one point :') put it back

apluslms_roman/configuration.py
```diff
@@ -56,16 +64,55 @@ def find_from(cls, path, name=None, prefixes=None):
...
+        super().initialize_data()
+        self.recursive_update(
+            LEGACY_CONFIG if is_legacy_configuration(self.dir)
+            else DEFAULT_CONFIG
+        )
+
+    def add_step(self, step):
+        if "name" in step:
+            other_names = [s["name"] for s in [s.get_data() for s in self.steps] if "name" in s]
+            if step["name"] in other_names:
```
raphendyr: I presume this can be replaced with `if step['name'] in self.steps_by_name`

apluslms_roman/settings.py
raphendyr: Can you test if this is still needed?

apluslms_roman/cli.py
ppessi: ....I fixed that at some point

apluslms_roman/cli.py
```diff
@@ -223,46 +226,93 @@ def add_cli_actions(parser):
...
+        help=_("actions for editing steps in the project configuration"))
+
+    with step.use_subparsers(title=_("Step actions")):
+        step.add_parser('list', aliases=['ls'],
+            callback=step_list_action,
+            help=_("list all steps and exit"))
+
+        add = step.add_parser('add', aliases=['a'],
+            callback=step_add_action,
+            help=_("add step to roman.yml"))
```
raphendyr: Maybe `add step to project config`?

apluslms_roman/cli.py
```diff
@@ -223,46 +226,93 @@ def add_cli_actions(parser):
...
+        add.add_argument('img')
+        add.add_argument('-n', '--name')
+        add.add_argument('-e', '--env', nargs='+', help=_("format: key=val"))
```
raphendyr, line 301: hmm.. how does `nargs='+'` or `nargs='*'` differ from `action='append'` in this kind of case? Documentation uses append for non-positional arguments and nargs for positional arguments in examples. Any case, `nargs='*'` probably should be used at least, so the `-e` wouldn't be required.

apluslms_roman/cli.py
```diff
@@ -223,46 +226,93 @@ def add_cli_actions(parser):
...
+        add.add_argument('-c', '--cmd')
+        add.add_argument('-m', '--mnt')
+
+        rm = step.add_parser('remove', aliases=['rm'],
+            callback=step_rm_action,
+            help=_("delete step from course.yml."))
```
raphendyr: Maybe `remove a step from project config` or at least from `roman.yml`?

apluslms_roman/cli.py
```diff
@@ -223,46 +226,93 @@ def add_cli_actions(parser):
...
+
     with validate.use_subparsers(title=_("Validate actions")):
         validate_schema = validate.add_parser('schema',
             callback=validate_schema_action,
             help=_("validate an YAML against a known schema"))
         validate_schema.add_argument('-v', '--version', metavar='version',
             dest='schema_version',
-            help=_("the version of a schema for validation (read from 'version' in a document by default)"))
+            help=_(
+                "the version of a schema for validation " +
```
raphendyr: you can drop the `+`.

```
>>> a = (
...  "foo"
...  "bar"
... )
>>> a
'foobar'
```

apluslms-yamlidator/apluslms_yamlidator/utils/collections.py
```diff
@@ -16,6 +16,19 @@
 else:
     from collections import OrderedDict
```
raphendyr, line 18: Two blank lines surrounding root level objects would be nice :)

apluslms-yamlidator/apluslms_yamlidator/validator.py
```diff
@@ -164,33 +162,46 @@ def __del__(self):
             f = getattr(self, key)
             logger.debug("%s: %s", key, f.cache_info())
```
raphendyr, line 164: Two blank lines surrounding root level objects would be nice :)

apluslms_roman/builder.py
```diff
@@ -20,16 +20,20 @@ def __init__(self, engine, config, observer=None):
...
-            try:
-                steps = [steps[ref] if isinstance(ref, int) else name_dict[ref] for ref in step_list]
-            except KeyError as e:
-                raise ValueError(_("No step named {}").format(e))
-            except IndexError as e:
-                raise ValueError(_("A step index was too big. Remember, indexing begins with 0."))
+            steps = []
+            for ref in refs:
+                try:
+                    steps.append(self.config.get_step(ref))
```
raphendyr: What do you think if this would be written:

```
get_step = self.config.get_step
try:
    steps = [get_step(ref) for ref in refs]
except KeyError:
    raise ValueError(_("No step named %s") % format(ref))
except IndexError:
    raise IndexError(_(
        "Index %s is out of range. There are %d steps. Indexing begins at 0."
    ) % (ref, len(self.config.steps)))
```

apluslms_roman/builder.py
```diff
@@ -60,11 +64,11 @@ def __init__(self, backend_class=None, settings=None):
 
         name = getattr(backend_class, 'name', None) or backend_class.__name__.lower()
         env_prefix = name.upper() + '_'
-        env = {k: v for k, v in environ.items() if k.startswith(env_prefix)}
+        env = {key: val for key, val in environ.items() if key.startswith(env_prefix)}
```
raphendyr: Even though this would be a good change, can you leave it as it's now. #8 is rewriting this part any case.

apluslms_roman/cli.py
```diff
@@ -67,10 +67,11 @@ def __init__(self, *args, callback_dest=None, callback_subparser_defaults=None,
 
         if callback_dest is None:
             callback_dest = 'action'
-        self.register('action', 'parsers', partial(CallbackSubParsersAction, callback_dest=callback_dest))
+        self.register('action', 'parsers',
+            partial(CallbackSubParsersAction, callback_dest=callback_dest))
         self._callback_dest = callback_dest
         self._callback_subparsers = None
-        self._callback_subparser_defaults = callback_subparser_defaults.copy() if callback_subparser_defaults is not None else {}
+        self._callback_subparser_defaults = (callback_subparser_defaults or {}).copy()
```
raphendyr: Old code is 50% faster (doesn't matter at all for single call :'D ), but I get that the old code is also bit hard to read. Suggestion:
```
if callback_subparser_defaults is not None:
  self._callback_subparser_defaults = callback_subparser_defaults.copy()
else:
  self._callback_subparser_defaults = {}
```

Your solution is nice short single line and my brain is competing against nice simple solution and obvious unnecessary work being done.

apluslms_roman/cli.py
```diff
@@ -167,13 +168,16 @@ def create_parser(version=__version__,
         metavar=_('DIR'),
         default=[],
         action='append',
-        help=_("Change to directory DIR before reading the configuration file or doing anything else. If specified multiple times, then values are joined with `os.path.join`."))
+        help=_(
+            "Change to directory DIR before reading the configuration " +
```
raphendyr: `+` is not required here. As a detail, '+' creates runtime code to calculate the addition, where as removing it tells compiler that here is a single string.

```
>>> pprint(parse("'foo' 'bar' 'baz'"))
Module(
    body=[
        Expr(
            lineno=1,
            col_offset=0,
            value=Str(lineno=1, col_offset=0, s='foobarbaz'),
        ),
    ],
)
>>> pprint(parse("'foo' + 'bar' + 'baz'"))
Module(
    body=[
        Expr(
            lineno=1,
            col_offset=0,
            value=BinOp(
                lineno=1,
                col_offset=14,
                left=BinOp(
                    lineno=1,
                    col_offset=0,
                    left=Str(lineno=1, col_offset=0, s='foo'),
                    op=Add(),
                    right=Str(lineno=1, col_offset=8, s='bar'),
                ),
                op=Add(),
                right=Str(lineno=1, col_offset=16, s='baz'),
            ),
        ),
    ],
)
```

apluslms_roman/cli.py
```diff
@@ -223,46 +226,93 @@ def add_cli_actions(parser):
         help=_("a list of sub commands:"))
     parser.use_subparsers(
         title=_("Commands"),
-        description=_("Top level commands. Check the per command help with `%(prog)s COMMAND --help`."))
+        description=_(
+            "Top level commands. Check the per command help " +
```
raphendyr: `+`, check other comments about the same

apluslms_roman/cli.py
```diff
@@ -342,20 +400,13 @@ def build_action(context):
...
-        print(header_fmt.format('ID', 'NAME', 'IAMGE'))
-        for step in steps:
-            print(step_fmt.format(step.ref, step.name or "", step.img))
-        return
-
     if not verify_engine(engine, only_when_error=True):
         return 1
 
+    # check that _build folder exists, create one if it doesn't
+    if not isdir('_build'):
```
raphendyr: If you move this just before https://github.com/apluslms/roman/blob/ea7bcdfdb0133a406bf4d27bbdae6e722b03832d/apluslms_roman/builder.py#L47, then I'll merge it in this PR. Though, add something like `# FIXME: add support for different build paths`.

apluslms_roman/cli.py
```diff
@@ -365,20 +416,136 @@ def build_action(context):
...
+        raise Exception("Choose either global settings or project settings")
+    document = context.settings if context.args.global_ else get_config(context)
+
+    for val in context.args.values:
+        try:
+            key, val = val.split('=', 1)
+            document.mlset_cast(key, val)
+        except ValueError as err:
+            if hasattr(err, 'value_type'):
+                raise ValueError(_("{} should be of type '{}', but was '{}'.").format(
```
raphendyr: We should avoid raising exceptions for the user. Basically, when the user doesn't benefit from the traceback, it will just fool them. Thus, prefer `exit(1, message)` here and below.

apluslms_roman/cli.py
```diff
@@ -365,20 +416,136 @@ def build_action(context):
...
+            else:
+                raise ValueError(_("Give values in format 'key=val'."))
+
+    document.validate()
+    document.save()
+
+
+def config_rm_action(context):
+    if context.args.global_ == context.args.project:
+        raise Exception("Choose either global settings or project settings")
```
raphendyr: Prefer `exit(1, message)` here too.

apluslms_roman/cli.py
```diff
@@ -365,20 +416,136 @@ def build_action(context):
...
+        print(step_fmt.format(i, step.get('name', ''), step['img']))
+    return
+
+
+def step_add_action(context):
+    args = context.args
+    env = args.env
+    if env:
+        try:
+            env = {a: b for a, b in [s.split('=', 1) for s in env]}
```
raphendyr: Format `env = {a: b for a, b in (s.split('=', 1) for s in env)}` would be nicer as it uses generator instead of a list. Though, you could also do `env = dict(s.split('=', 1) for s in env)` or even better, `env = OrderedDict(s.split('=', 1) for s in env)`.

apluslms_roman/cli.py
```diff
@@ -365,20 +416,136 @@ def build_action(context):
...
+        except ValueError:
+            raise ValueError("env is a dict, so values need to be in key=val format, e.g. a=1 b=2")
```
raphendyr: The same `exit(1, message)` thing.

apluslms_roman/configuration.py
```diff
@@ -56,16 +64,54 @@ def find_from(cls, path, name=None, prefixes=None):
...
+        self.recursive_update(
+            LEGACY_CONFIG if is_legacy_configuration(self.dir)
+            else DEFAULT_CONFIG
+        )
+
+    def add_step(self, step):
+        if 'name' in step:
+            if step['name'].lower() in self.steps_by_name:
+                raise ValueError("A step with the name '%s' already exists." % step['name'])
+        self.steps.append(step)
```
raphendyr, line 66: You probably need to ensure `steps` is in the self container. In case where `steps` is not in the container, then we would effectively do `[].append(step)`. Of course, this is mostly handled by schema (for now), so this shouldn't ever happen. Though, schema probably shouldn't require at least one item in steps list.

```
if 'steps' not in self:
  self['steps'] = [step]
else:
  self['steps'].append(step)
```

It's possible that `self.setdefault('steps', [])` in `def steps(self):` would be a better solution and it might even work correctly. Would set 'steps' to the ChangesDict default set and when an value is added to the list, it should move the list from defaults to data (at least ChangesList should work that way if it doesn't currently).

apluslms_roman/schemas/roman_project-v2.0.yaml
```diff
@@ -3,39 +3,45 @@ $schema: "http://json-schema.org/draft-04/schema"
...
-
 allOf:
   - $ref: version-v1.0
-
 additionalProperties: false
 properties:
   version: {}
   steps:
     type: array
+    minItems: 1
```
raphendyr: Based on commit https://github.com/apluslms/roman/pull/9/files#r294316902, we need to revisit this idea.

apluslms_roman/builder.py
ppessi: doesn't work like that, python doesn't remember ref outside of the list comprehension

apluslms_roman/cli.py
```diff
@@ -365,20 +416,136 @@ def build_action(context):
...
+    step = {
+        'img': args.img,
+        'cmd': args.cmd,
+        'mnt': args.mnt,
+        'env': env,
+        'name': args.name
+    }
+    step = {k: v for k, v in step.items() if v}
+    config = get_config(context)
+    config.add_step(step)
```
raphendyr: Not sure if `config.add_step` should use definition `def add_step(img, cmd=None, mnt=None, env=None, name=None)` instead. This would require to call it `config.add_step(**step)`, but then the python interface would ensure that only valid values are added to a step. Sadly, then `add_step` needs to filter `None` or false values on that side too.

Any thoughts?

apluslms_roman/cli.py
```diff
@@ -365,20 +416,136 @@ def build_action(context):
...
+        step = config.get_step(ref)
+    except IndexError:
+        raise IndexError(_(
+            "Index is out of range. Remember, indexing start from 0. " +
+            "Use 'roman --list-steps' to see step indexes."))
+    except KeyError:
+        raise KeyError(_("There is no step called '{}'").format(ref))
+    if not context.args.force and not confirm_del(step):
+        return
+    config.del_step(ref)
```
raphendyr: It would be nice if the step dict could be passed to `del_step` too (string/int ref is also important interface). Current solution is fine for now. Just thinking aloud,

apluslms_roman/builder.py
raphendyr: Damn, didn't notice `ref` in `raise ValueError(_("No step named %s") % format(ref))`. My bad.

apluslms_roman/configuration.py
ppessi: should I change schema to not require steps? although what is the point of a course config if it doesn't have steps in it?

apluslms_roman/cli.py
ppessi: I don't really understand what you mean with "then the python interface would ensure that only valid values are added to a step"...

apluslms_roman/cli.py
ppessi: well I changed `del_step` so it allows dicts too, didn't really require much. (oops forgot to send this comment earlier)

apluslms_roman/cli.py
ppessi: The `nargs='+'` doesn't mean that `--env` is required, it means that if env is used, at least one argument is required, which is how we want it to work. Because `--env` has `--` in front of it, it isn't a required argument. `action='append'` means that for each `-e` there needs to be exactly one argument, so if the user wants to put several values into the env, they need to write the flag several times.

raphendyr: Hahaa. Great! Then it sounds exactly the correct. Yey. :)

apluslms_roman/configuration.py
raphendyr: Nothing, except it would tell roman that this is a course/project folder, so it doesn't raise error (no config found). The legacy config tries to do sphinx build and there are courses that use only yaml files.

Later on, there might be service definitions, which might be the only thing required in that case. But yeah, currently it would seem silly to have config without any steps.

That said, I guess the `steps` key shouldn't be required either.

apluslms_roman/cli.py
raphendyr: `add_steps(**a_dict)` -> `def add_step(img, cmd=None, mnt=None, env=None, name=None)` conversion checks that there are only keys `img`, `cmd`, `mnt`, `env` and `name`. Any other key would raise error.

Alternative would be for `add_step` to validate that the dictionary doesn't contain any invalid keys.

Though, now that I think about it, this might introduce unnesesary duplication of logic as schema ensures the dictionary is valid in the end.

If the GUI benefits from a different type of interface, then someone can add that.

Thus, after rethinking, let's go with what there is now.

apluslms_roman/cli.py
```diff
@@ -334,11 +341,18 @@ def get_config(context):
         if context.args.project_config:
             project_config = abspath(expanduser(expandvars(context.args.project_config)))
             return CourseConfig.load_from(project_config)
-        return CourseConfig.find_from(getcwd())
+        try:
+            return CourseConfig.find_from(getcwd())
+        # getcwd() is always a valid path, so only the lack of a roman config
+        # file can cause a CourseConfigError
+        except CourseConfigError as err:
+            raise FileNotFoundError(str(err) +
```
ppessi: init_action catches this, so exit can't be used because even if SystemExit is catched, it prints the error.

apluslms_roman/cli.py
```diff
@@ -365,20 +418,136 @@ def build_action(context):
...
+            document.mlset_cast(key, val)
+        except ValueError as err:
+            if hasattr(err, 'value_type'):
+                exit(1, _("{} should be of type '{}', but was '{}'.").format(
+                    key, err.value_type, type(val).__name__))
+            else:
+                exit(1, _("Give values in format 'key=val'."))
+
+    document.validate()
+    document.save()
```
ppessi: ....this doesn't the same way it does when loading a file?

apluslms_roman/cli.py
raphendyr: We need to revisit the design for interfaces of ProjectConfig, get_config and init_action. I think we need to find a solution where get_config and init_action are separate, but still simple (e.g. no code duplication).

For example, I think we need FileNotFoundError in ProjectConfig, so we can see when missing file is the problem and when the file itself is the problem.

apluslms_roman/cli.py
raphendyr: I created some tests for different things, but didn't find any reason for the problem so far. I'll debug with the PR code tomorrow.

apluslms_roman/cli.py
```diff
@@ -195,17 +207,20 @@ def parse_actioncontext(parser):
 
     # load settings
     allow_missing = args.config is None
```
raphendyr: This variable isn't used anymore.

apluslms_roman/cli.py
```diff
@@ -195,17 +207,20 @@ def parse_actioncontext(parser):
...
-        settings = RomanSettings.load(config, allow_missing=allow_missing)
+        settings = GlobalSettings.load(config, allow_missing=True)
     except ValidationError as e:
         exit(1, '\n'.join(render_error(e)))
-    except IOError as e:
+    except OSError as e:
         exit(1, str(e))
 
+    if args.config is not None and not settings.container.exists():
+        warning("File {} doesn't exist.".format(args.config))
```
raphendyr: In same update, you could wrap the string in `_()`.

apluslms_roman/cli.py
```diff
@@ -365,20 +435,155 @@ def build_action(context):
...
+        exit(1, _("A project configuration already exists at {}".format(config.path)))
+    except FileNotFoundError:
+        pass
+
+    if context.args.project_config:
+        file_ = context.args.project_config
+        if '.' not in file_:
+            is_recognized = False
+        else:
+            name, prefix = file_.split('.', 1)
```
raphendyr: This `name, _delim, prefix = file_.rpartition('.')` would support filenames `foo.bar.yml`. Of course, `foo.bar` is not in DEFAULT_NAMES, thus following test would fail anycase. Just pointing out that when working with filename extensions `rpartition`, `rsplit` and `os.path.splitext`.

To note, current implementation does work, even though `name` and `prefix` will contain strange values in many cases when we get value `is_recognized = False`.

apluslms_roman/cli.py
```diff
@@ -365,20 +435,155 @@ def build_action(context):
...
+            if (name not in ProjectConfig.DEFAULT_NAMES
+                    or prefix not in ProjectConfig.DEFAULT_PREFIXES):
+                is_recognized = False
+
+        if not is_recognized:
```
raphendyr, line 468: With `roman -f roman.yml` this should raise `is_recognized` is not initialized. None of the code sets it to True.

apluslms_roman/cli.py
```diff
@@ -365,20 +435,155 @@ def build_action(context):
...
         print("---\n# roman settings:")
         yaml_dump(context.settings._data, stdout)
     if all_ or context.args.project:
         print("---\n# project config")
         yaml_dump(get_config(context)._data, stdout)
 
 
+def config_set_action(context):
+    if context.args.global_ == context.args.project:
+        exit(1, "Choose either global settings or project settings")
```
raphendyr, line 496: If neither are set, then default to project config, so `roman config set foo=bar` uses project config. If user would say `roman config -g -p` then we need to raise error. `-p` is mostly for someone to be explicit (e.g. to ensure we write project config in a script).

apluslms_roman/tests/test_cli.py
```diff
@@ -0,0 +1,285 @@
...
+    old_err = (sys.stderr, cli.stderr)
+    try:
+        sys.stdout = cli.stdout = new_out
+        sys.stderr = cli.stderr = new_err
+        yield (sys.stdout), (sys.stderr)
+    finally:
+        sys.stdout, sys.stderr = old_out[0], old_err[0]
+        cli.stdout, cli.stderr = old_out[1], old_err[1]
+
+COURSE = join(getcwd(), 'mock_course')
```
raphendyr: `python3 setup.py test` should fail with `getcwd`. You should use `join(dirname(__file__), 'mock_course')`

apluslms_roman/tests/test_cli.py
```diff
@@ -0,0 +1,285 @@
...
+        if isdir(file_):
+            rmtree(file_)
+        else:
+            remove(file_)
+
+def use_temp_file(func):
+    def wrapper(*args):
+        delete('temp.yml')
+        command_test(args[0], "roman -f temp.yml init")
+        try:
```
raphendyr:
```
try:
  func(*args)
finally:
  delete(..)
```

apluslms_roman/tests/test_cli.py
```diff
@@ -0,0 +1,285 @@
...
+CONFIG = 'roman.yml'
+
+def delete(file_):
+    if exists(file_):
+        if isdir(file_):
+            rmtree(file_)
+        else:
+            remove(file_)
+
+def use_temp_file(func):
```
raphendyr: This probably could be implemented with `tempfile.NamedTemporaryFile`

apluslms_roman/tests/test_cli.py
```diff
@@ -0,0 +1,285 @@
...
+        command_test(args[0], "roman -f temp.yml step add hello-world --name hello")
+        func(*args)
+    return wrapper
+
+def format_output(io_string):
+    return io_string.getvalue().strip()
+
+def command_test(cls, command, exit_code=None):
+    if isinstance(command, str):
+        command = command.split(" ")
```
raphendyr: https://docs.python.org/3/library/shlex.html#shlex.split

apluslms_roman/tests/test_cli.py
```diff
@@ -0,0 +1,285 @@
...
+    sys.argv = command
+    exit_code = str(exit_code) if exit_code is not None else '0'
+    try:
+        with capture_output() as (out, err):
+            cls.assertRaisesRegex(SystemExit, exit_code, cli.main)
```
raphendyr: `cli.main` could take `argv` as an argument and if None is given, then it would use `sys.argv`.

apluslms_roman/tests/test_cli.py
ppessi: I did check at one point how that works but apparently forgot :p

apluslms_roman/tests/test_cli.py
ppessi: `python3 cli.py test` doesn't work anyway because of the relative imports :cry:

apluslms_roman/tests/test_cli.py
```diff
@@ -0,0 +1,278 @@
...
+        _, err = command_test(self, "roman -f temp.yml step rm -f 2", 1)
+        self.assertIn("Index is out of range", err)
+
+    @use_temp_file
+    @add_hello
+    def test_rm_action_key_error(self):
+        _, err = command_test(self, "roman -f temp.yml step rm -f hei", 1)
+        self.assertEqual("There is no step called 'hei'", err)
+
+chdir(COURSE)
```
raphendyr: We shouldn't change current dir, as that might do unintended side effects. A bit better is to do someting like: `command_test(self, "roman -f '%s' step rm -f hei" % (join(COURSE, 'temp.yml'), 1)`

Btw. I would implement a parent class like
```
class Foo(unitetest.TestCase):
  def command_test(self):
    pass # implementation here

class Test1(Foo):
  def test_rm_action_key_error(self):
    _, err = self.command_test("roman -f plaa")
```

`command_test` could even have pattern `def command_test(config_filename, subcommand)` and would then do something like `cmd = ['roman', '-f', config_filename] + shlex.split(subcommand)`

apluslms_roman/tests/test_cli.py
raphendyr: It should. At least the test in travis-ci work for current tests in milestone0.3 and feature/new_observer.

ppessi: I tried calling from a few different folders and it didn't work. Maybe moving the test to the root of the project could help? There's at least something there. Will try at some point.

apluslms_roman/tests/test_cli.py
ppessi: Yeah that would look a lot better... Still not very used to doing class based programming. :p

apluslms_roman/tests/test_cli.py
ppessi: so `args = parser.parse_args()` is the part that should be changed?

apluslms_roman/cli.py
ppessi: Yeah I did consider that there might be several periods in the name, but since it wouldn't be any of the default names I decided to ignore it. Good to know about the filename methods though.

apluslms_roman/tests/test_cli.py
ppessi: Moving them to the tests folder at root helped :) Should I move yamlidator tests there, too?
EDIT: it didn't, I remembered the command wrong... `python3 cli.py` doesn't work either, if that info is somehow useful.

raphendyr: Roman client can be executed with `python3 -m apluslms_roman.cli`. I pushed small commit to milestone0.3, which allows to use `python3 -m apluslms_roman`.

apluslms_roman/configuration.py
```diff
@@ -55,17 +42,48 @@ def find_from(cls, path, name=None, prefixes=None):
...
+                return self.steps[idx]
+            except IndexError as err:
+                # IndexError doesn't include the index by default
+                raise IndexError(idx) from err
+        return self.steps_by_name[step_ref.lower()]
 
     @property
     def steps(self):
-        return self.get('steps') or []
+        return self.mlsetdefault('steps', [])
```
raphendyr: `self.setdefault('steps', [])` should be enough

tests/test_cli.py
```diff
@@ -0,0 +1,410 @@
...
+                return self.CommandResult(out.getvalue(), err.getvalue(), vfs)
+            except Exception as e:
+                if exit_mock.called:
+                    exit_args = exit_mock.call_args[0]
+                    returned_code = exit_args and exit_args[0] or 0
+                else:
+                    returned_code = '?'
+                print("Error traceback:", file=err)
+                print(str(e), file=err)
+                print_tb(e.__traceback__, limit=10, file=err)
```
raphendyr: It's bit questionable to print test output to `err`, which is the stderr of the `cli:main` execution. In theory, `raise AssertionError(...) from e` should be enough, but I get that the `unittest` framework seems to eat some of the important traceback lines.

How about something like?:

```
raise AssertionError(
    "cli.main({!r}) -> {} raised an exception\n\n{}\n"
    "Roman output:\n{}\n{}"
    .format(
        args,
        returned_code,
        format_exc(),
        out.getvalue().strip(),
        err.getvalue().strip(),
    ))
```

It seems `format_exc()` renders a string with all of the exception details.

raphendyr: The original problem with `unittest` is that it probably uses `print_exc(limit=2)`. The positive limit means, that 2 frames from our current location forward and thus doesn't show the line which actually raised the error. I would presume `limit=-10` would be a better default to avoid this.

The best would be to show 5 frames up from current location and -3 last frames (presuming there is more than 8 frames) for the latest exception and -2 for all previous errors in the chain. Implementing that here would be waste of a time though.

tests/test_cli.py
```diff
@@ -0,0 +1,410 @@
...
+        r = self.command_test("step list", config={'version': '2.0'})
+        self.assertEqual("The project config has no steps.", r.out.strip())
+
+
+class TestStepAddAction(CliTestCase):
+
+    def test_step_add(self):
+        r = self.command_test(
+            "step add hello-world --name new_step "
+            "--cmd \"make touchrst html\" "
```
raphendyr: You should use different quotes on outer and inner strings. e.g. `"-- cmd 'make touchrst html' "`.

raphendyr:
Look into https://github.com/apluslms/roman/pull/9/files#r298632787, rebase and then this is ready for merging.

tests/test_cli.py
ppessi: If I recall correctly I couldn't find a method for turning the error into a traceback string without printing, and I would've jsut printed the info but the mocked out and err ate it so I ended with that. :p

ppessi: Apparently all I had to do was scroll the docs down a bit... oh well

raphendyr:
Yey. One step closer to v0.3 release :slightly_smiling_face:

raphendyr:
p.s. there were some problem in yamlidator tests, but fixed those in milestone0.3 after the merge.

raphendyr:
PR included some `*.pyc`. I rewrite the history, so the last commit in repo is actually 0ce3bf0b6144fad7c2f73baada724d531b78eccb and not the one mentioned by github.

PR #11, opened 2019-07-05
Reorder cli commands: fix step printing

ppessi:
Earlier I had deleted the stuff from build_action which enabled step listing with `roman build --steps ?`, now I added step listing back to build_action.


raphendyr:
PR #9 included some `*.pyc`. I rewrite the history, so the commit in repo is actually 51ad312701c9a2e73157ff6486e8d7cbbe028a9c and not the one mentioned by github.

PR #12, opened 2019-07-05
fix get_steps

ppessi:
Fixed get_steps in BuildStep


tests/test_builder.py
```diff
@@ -0,0 +1,37 @@
...
+            ]}
+        config = ProjectConfig(ProjectConfig.Container(
+            '/a', allow_missing=True), None, config, None)
+        self.builder = Builder(None, config)
+
+    def test_get_steps(self):
+        steps = self.builder.get_steps()
+        self.assertEqual(len(steps), 3)
+        names = [s.name for s in steps]
+        self.assertEqual(['test1', 'test2', None], names)
```
raphendyr: It's might be good idea to switch actual and expected parameters around, just to keep consistency between tests.

raphendyr:
PR #9 included some `*.pyc`. I rewrite the history, so the commit in repo is actually 43e2f77dd3695d654072bc281e743e8e41a93d7c and not the one mentioned by github.

PR #13, opened 2019-07-30
Validation stuff and clean build

ppessi:
Improved validation errors, ChangesList has separate work and data so lc can now be used with it.


raphendyr:
Please rebase to fix the conflict. This is good for merging after that.

PR #16, opened 2019-07-31
Env stuff

ppessi:


apluslms-yamlidator/apluslms_yamlidator/utils/collections.py
```diff
@@ -93,44 +93,53 @@ def data_updated(self, key, data):
     def get_data(self):
         return self._data
 
+    @property
+    def top_parent(self):
+        if not self._parent:
+            return self
+        parent = self._parent
+        if not parent:
+            return self
```
raphendyr, line 99: :thinking: `if not self._parent and not self._parent`, I think you have the same tests twice.

apluslms-yamlidator/apluslms_yamlidator/utils/error_render.py
```diff
@@ -0,0 +1,22 @@
+def print_lc(file_, line, col, num_lines=5, print_file=True):
```
raphendyr: As the output is list of strings (or if it was just a string), then a name `render` is more suited. `print` indicates the funcion doesn't output anything.

raphendyr: Or it could be `format` similar to functions in traceback module.

apluslms-yamlidator/apluslms_yamlidator/document.py
```diff
@@ -409,7 +408,13 @@ def mlget(self, keys, default=_NoDefault):
 
     def mlset(self, keys, value):
         container, key = find_ml(self._data, keys, create_dicts=True)
-        container[key] = value
+        if isinstance(container, (list, Sequence)) and not isinstance(key, int):
```
raphendyr: This is also done in `find_ml`?

apluslms-yamlidator/apluslms_yamlidator/utils/collections.py
```diff
@@ -93,44 +93,51 @@ def data_updated(self, key, data):
     def get_data(self):
         return self._data
 
+    @property
+    def top_parent(self):
```
raphendyr: This is typically called `root`. As `_parent` is non-public, this could be too.

tests/test_builder.py
```diff
@@ -18,18 +18,18 @@ def setUp(self):
...
 
-    def test_get_steps(self):
+    def test_basic(self):
         steps = self.builder.get_steps()
         self.assertEqual(len(steps), 3)
         names = [s.name for s in steps]
         self.assertEqual(names, ['test1', 'test2', None])
 
-    def test_get_steps_with_ref_duplicate(self):
+    def test_ref_duplicate(self):
```
raphendyr: Longer test names typically explain more. Something to consider: https://qualitycoding.org/unit-test-names/ . I used another schema in https://github.com/Aalto-LeTech/a-plus/blob/master/shibboleth_login/tests.py and it seems to create relatively readable names.

apluslms_roman/utils/env.py
```diff
@@ -0,0 +1,275 @@
...
+
+def get_val(env, key, value):
+    references_self = _("Variable {} references itself")
+    not_defined = _("{} hasn't been defined")
+
+    if isinstance(value, int):
+        return value
+    value = to_str(value)
+    while True:
+        var = re.search(sub_rgx, value)
```
raphendyr, line 140: Yep. This is the place that would be nicer with the assignment expressions https://www.python.org/dev/peps/pep-0572/, sadly we can't use it yet :(

raphendyr:
Some things to consider, but I think this is good for merging after rabase.

apluslms-yamlidator/apluslms_yamlidator/utils/collections.py
ppessi: Ah yes thank you, I had trouble figuring a good name :p

tests/test_builder.py
ppessi: well I wouldn't want there to be tests with different naming conventions, so... should I rename them all? I just thought the class being called 'TestBuilder' was kind of inaccurate and changed it to 'TestGetSteps', and then mentioning 'get_steps' in the function names would've been redundant.

raphendyr: You could use `TestBuilderGetSteps` to include class and method, and then not include the method name in test function. Though,  you would like to include the other two parts in it (what is given and what is supposed to happen).

raphendyr: You can rename them all, or start using the concept for all future tests.

ppessi: Well I renamed all of them, but I'm not sure how much better the new names are...

apluslms-yamlidator/apluslms_yamlidator/utils/collections.py
```diff
@@ -93,44 +93,54 @@ def data_updated(self, key, data):
     def get_data(self):
         return self._data
 
+    def get_root(self):
+        return self._root
+
+    @property
+    def _root(self):
```
raphendyr: I'm not sure if `_root` is usefull here, as `get_root` is the only one used. Alternatively, last line of this property should probably have line `parent._root` and not `parent.get_root()`, Currently, the flow of the code jumps a bit around.

PR #19, opened 2019-08-06
Support for multiple configs for same backend class

ppessi:
https://trello.com/c/GkFxBBMK/24-support-for-multiple-configurations-for-same-backend-class

### DIFF
--- a/apluslms_roman/builder.py
+++ b/apluslms_roman/builder.py
@@ -55,26 +55,42 @@ class Builder:
         return result
 
 
+class BackendError(Exception):
+
+    def __init__(self, backend):
+        super().__init__()
+        self.backend = backend
+
+
 class Engine:
     def __init__(self, backend_class=None, settings=None):
+        backend_name = None
         if backend_class is None:
             if settings and 'backend' in settings:
                 backend_class = settings['backend']
+                if 'backends' in settings and backend_class in settings['backends']:
+                    backend_name = backend_class
+                    if 'type' in settings['backends'][backend_name]:
+                        backend_class = settings['backends'][backend_name]['type']
             else:
                 from .backends.docker import DockerBackend as backend_class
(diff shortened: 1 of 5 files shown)


apluslms_roman/builder.py
```diff
@@ -51,24 +51,39 @@ def build(self, step_refs: list = None):
...
-        if settings:
-            for k, v in settings.get(name, {}).items():
-                if v is not None and v != '':
-                    env[env_prefix + k.replace('-', '_').upper()] = v
+        if backend_name is not None:
+            env.update({env_prefix + k.replace('-', '_').upper(): v
+                for k, v in settings['backends'][backend_name].items()
+                if k != 'type'})
+
+        if backend_class in DEFAULTS:
```
raphendyr: Let's drop this. When we actually need for it, let's reimplement.

apluslms_roman/builder.py
```diff
@@ -51,24 +51,39 @@ def build(self, step_refs: list = None):
...
             else:
                 from .backends.docker import DockerBackend as backend_class
         if isinstance(backend_class, str):
             if '.' not in backend_class:
                 backend_class = BACKENDS.get(backend_class, backend_class)
-            backend_class = import_string(backend_class)
+            try:
+                backend_class = import_string(backend_class)
+            except ImportError as err:
+                err.backend = backend_class
```
raphendyr: It would be better to raise a different error, so only it would be captured in cli and thus any other importerror would not be captured by the cli.

ppessi: hm well cli could also raise the error if it doesn't have the backend attribute?

raphendyr: I think it's cleaner to use different error class, than adding a variable to it.

ppessi:
Travis checks are failing because #26 is needed for the tests to work.

PR #17, opened 2019-08-02
Coloured build output

ppessi:
`--no-color` flag can be used to turn colors off

PR #20, opened 2019-08-07
Colour output

ppessi:
basically pr #17 and #18 combined

### DIFF
--- a/apluslms_roman/backends/__init__.py
+++ b/apluslms_roman/backends/__init__.py
@@ -67,23 +67,39 @@ class BuildStep:
 
 
 class BuildResult:
-    __slots__ = ('code', 'error', 'step')
+    __slots__ = ('ok', 'code', 'error', 'step')
 
-    def __init__(self, code=0, error=None, step=None):
+    def __init__(self, ok=None, *, code=0, error=None, step=None):
+        if error is not None and code == 0:
+            code = -1
+        self.ok = ok if ok is not None else code == 0
         self.code = code
         self.error = error
         self.step = step
-        assert self.ok or step is not None, "step is required for failed result"
+        # it is possible for build to be cancelled without knowing
+        # which step was in progress, so we can't require step
+        # for cancelled build
+        assert code == 0 or step is not None, "step is required for failed result"
 
     @property
-    def ok(self):
-        return self.code == 0 and self.error is None
+    def cancelled(self):
+        return not self.ok and self.code == 0
+
+    @property
+    def failed(self):
+        return not self.ok and self.code != 0
 
     def __str__(self):
         if self.ok:
             return "Build ok"
+        if not self.step:
+            return "Build cancelled"
+        status = 'cancelled' if self.cancelled else 'failed'
+        msg = "Build {} on step {}".format(status, self.step)
+        if self.cancelled:
+            return msg
         error = self.error or 'exit code {}'.format(self.code)
-        return "Build failed on step {}: {}".format(self.step, error)
+        return "{}: {}".format(msg, error)
 
 
 Environment = namedtuple('Environment', [
--- a/apluslms_roman/backends/docker.py
+++ b/apluslms_roman/backends/docker.py
@@ -88,20 +88,24 @@ You might be able to add yourself to that group with 'sudo adduser docker'.""")
     def prepare(self, task, observer):
         client = self._client
         for step in task.steps:
-            observer.step_preflight(step)
-            image, tag = step.img.split(':', 1)
             try:
-                img = client.images.get(step.img)
-            except docker.errors.ImageNotFound:
-                observer.step_running(step)
-                observer.manager_msg(step, "Downloading image {}".format(step.img))
+                observer.step_preflight(step)
+                image, tag = step.img.split(':', 1)
                 try:
-                    img = client.images.pull(image, tag)
-                except docker.errors.APIError as err:
-                    observer.step_failed(step)
-                    error = "%s %s" % (err.__class__.__name__, err)
-                    return BuildResult(-1, error, step)
-            observer.step_succeeded(step)
+                    img = client.images.get(step.img)
+                except docker.errors.ImageNotFound:
+                    observer.step_running(step)
+                    observer.manager_msg(step, "Downloading image {}".format(step.img))
+                    try:
+                        img = client.images.pull(image, tag)
+                    except docker.errors.APIError as err:
+                        observer.step_failed(step)
+                        error = "%s %s" % (err.__class__.__name__, err)
+                        return BuildResult(code=-1, error=error, step=step)
+                observer.step_succeeded(step)
+            except KeyboardInterrupt:
+                observer.step_cancelled(step)
+                return BuildResult(False, step=step)
         return BuildResult()
 
     def build(self, task, observer):
@@ -109,7 +113,7 @@ You might be able to add yourself to that group with 'sudo adduser docker'.""")
         for step in task.steps:
             observer.step_pending(step)
             opts = self._run_opts(task, step)
-            observer.manager_msg(step, "Starting container {}:".format(opts['image']))
+            observer.manager_msg(step, "Starting container {}".format(opts['image']))
             try:
                 with create_container(client, **opts) as container:
                     observer.step_running(step)
@@ -119,16 +123,16 @@ You might be able to add yourself to that group with 'sudo adduser docker'.""")
             except docker.errors.APIError as err:
                 observer.step_failed(step)
                 error = "%s %s" % (err.__class__.__name__, err)
-                return BuildResult(-1, error, step)
+                return BuildResult(code=-1, error=error, step=step)
             except KeyboardInterrupt:
                 observer.step_cancelled(step)
-                raise
+                return BuildResult(False, step=step)
             else:
                 code = ret.get('StatusCode', None)
                 error = ret.get('Error', None)
                 if code or error:
                     observer.step_failed(step)
-                    return BuildResult(code, error, step)
+                    return BuildResult(code=code, error=error, step=step)
                 observer.step_succeeded(step)
         return BuildResult()
 
--- a/apluslms_roman/builder.py
+++ b/apluslms_roman/builder.py
@@ -5,7 +5,7 @@ from shutil import rmtree
 from apluslms_yamlidator.utils.decorator import cached_property
 from apluslms_yamlidator.utils.collections import OrderedDict
 
-from .backends import BACKENDS, BuildTask, BuildStep, Environment
+from .backends import BACKENDS, BuildResult, BuildTask, BuildStep, Environment
 from .observer import StreamObserver
 from .utils.importing import import_string
 from .utils.translation import _
@@ -36,22 +36,24 @@ class Builder:
         backend = self._engine.backend
         observer = self._observer
         steps = self.get_steps(step_refs) # NOTE: may raise KeyError or IndexError
-
-        task = BuildTask(self.path, steps)
-        observer.enter_prepare()
-        result = backend.prepare(task, observer)
-        observer.result_msg(result)
-        if result.ok:
-            observer.enter_build()
-            # FIXME: add support for other build paths
-            if clean_build:
-                if isdir('_build'):
-                    rmtree('_build')
-            if clean_build or not isdir('_build'):
-                mkdir('_build')
-            result = backend.build(task, observer)
+        try:
+            task = BuildTask(self.path, steps)
+            observer.enter_prepare()
+            result = backend.prepare(task, observer)
             observer.result_msg(result)
-        observer.done(result)
+            if result.ok:
+                observer.enter_build()
+                # FIXME: add support for other build paths
+                if clean_build:
+                    if isdir('_build'):
+                        rmtree('_build')
+                if clean_build or not isdir('_build'):
+                    mkdir('_build')
+                result = backend.build(task, observer)
+                observer.result_msg(result)
+            observer.done(result)
+        except KeyboardInterrupt:
+            return BuildResult(False)
         return result
 
 
--- a/apluslms_roman/observer.py
+++ b/apluslms_roman/observer.py
@@ -1,6 +1,8 @@
 import sys
 from enum import Enum
+from time import time
 
+from colorama import init as init_color, Fore, Style
 
 class Phase(Enum):
     NONE = 0
@@ -104,7 +106,8 @@ class BuildObserver:
             raise RuntimeError(
                 "%s has not entered any phase when requested to update state to %s for step %r"
                 % (self.__class__.__name__, state, step))
-        if self.get_step_state(step) != state:
+        cur_state = self.get_step_state(step)
+        if cur_state != state and not cur_state.completed:
             self._states[step] = state
             self._message(self._phase, Message.STATE_UPDATE, step, state)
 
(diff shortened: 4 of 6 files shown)


ppessi:
Do the Travis CI tests test build output and fail because it's different from before or have I broken something?

raphendyr:
> Do the Travis CI tests test build output and fail because it's different from before or have I broken something?

You can check the travis log behind the details link. I presume it's visible for you too (should be). From the travis log, we can see that you have not included `colorama` in the requirements.txt.

> ModuleNotFoundError: No module named 'colorama'

apluslms_roman/backends/docker.py
```diff
@@ -111,7 +112,7 @@ def build(self, task, observer):
                 return BuildResult(-1, error, step)
             except KeyboardInterrupt:
                 observer.step_cancelled(step)
-                raise
+                exit()
```
raphendyr: Can't call exit here, as then a shepherd task worker would quit :)

apluslms_roman/observer.py
```diff
@@ -145,35 +147,77 @@ def result_msg(self, result):
...
-        if type_ == Message.STATE_UPDATE:
-            data = ENTER_STATE_TEXTS.get(state)
-            if data is None:
+        def print_time(duration, succeeded=True):
+            duration = int(duration)
+            if succeeded:
+                self.write('ok', Fore.GREEN + Style.BRIGHT)
+            else:
+                self.write('failed', Fore.RED + Style.BRIGHT)
+            hours, seconds = divmod(duration, 3600)
```
raphendyr, line 192: The time renderding could be a helper function, and then you can do stuff like `self.write("%s: %s" % (status_message, format_time(duration)), status_color)`

raphendyr: Hmm.. Though the time is not colored... well any case.

requirements.txt
```diff
@@ -1,2 +1,3 @@
 apluslms-yamlidator
 appdirs >=1.4.0, <2
+colorama >=0.4.1
```
raphendyr: `colorama ~= 0.4.1` would be preferred (appdirs should be updated to use that format some day..).

apluslms_roman/backends/docker.py
```diff
@@ -111,7 +111,7 @@ def build(self, task, observer):
                 return BuildResult(-1, error, step)
             except KeyboardInterrupt:
                 observer.step_cancelled(step)
-                raise
+                return BuildResult(step=step)
```
raphendyr: I think we should include the cancellation info in the buildresult object.

Thus, I think new BuldResult would include following:

| "state" | .ok | .cancelled | .failed | .code |
| ---------- | ---------- | ------- | -------- | --- |
| SUCCEEDED | True | False | Fasle | 0 |
| CANCELLED | False | True | False | 0 |
| FAILED | Fase | False | True | != 0 |

`.ok` needs to be changed from property to a variable and these could be properties `.cancelled = not .ok and .code == 0` and `.failed = not .ok and .code != 0`.

After that, we should change `Observer.result_msg` to include `_state_update(step, state)` based on the value in `result` object. This would trigger step state update, if it was not done yet (should have been done, but in theory). This ensure that the state send wiht `RESULT_MSG` is always correct.

ppessi: Is cancelled a necessary property? If the step neither succeeded or failed we know it was cancelled.

apluslms_roman/observer.py
```diff
@@ -151,38 +153,81 @@ def container_msg(self, step, msg):
         self._send_message(Message.CONTAINER_MSG, step, msg)
 
     def result_msg(self, result):
+        state = (StepState.SUCCEEDED if result.ok
+            else (StepState.CANCELLED if result.cancelled
+                else StepState.FAILED))
+        self._state_update(result.step, state)
```
raphendyr: I was thinking something like this:

```
    def result_msg(self, result):
        state = (StepState.SUCCEEDED if result.ok else
                 StepState.CANCELLED if result.cancelled else
                 StepState.FAILED)
        steps = (result.step,) if result.step else self._states
        for step in steps:
            self._state_update(step, state)
        self._message(self._phase, Message.RESULT_MSG, result.step, state, (result.code, result.error))
```
Presume that for `SUCCEEDED`, there is no step given (as all steps must succeed for result to be ok). Thus, what we can do (but shouldn't be mandatory), is to ensure all steps are marked (and notified) to be in `SUCCEEDED` state. After that, we will send result msg (directly as state would be `UNKNOWN` for None state).

This should make it possible for result message to have optional step.

apluslms_roman/backends/__init__.py
```diff
@@ -67,21 +67,25 @@ def __str__(self):
...
 class BuildResult:
-    __slots__ = ('code', 'error', 'step')
+    __slots__ = ('step', 'ok', 'code', 'error')
 
-    def __init__(self, code=0, error=None, step=None):
-        self.code = code
-        self.error = error
-        self.step = step
-        assert self.ok or step is not None, "step is required for failed result"
+    def __init__(self, step, ok=True, code=0, error=None):
```
raphendyr: Step is not very logical for `ok` result, as its not specific step where the build succeeded. But when build is stopeed before the end, we would like to know where it was stopped (and why).

apluslms_roman/backends/__init__.py
```diff
@@ -67,21 +67,25 @@ def __str__(self):
...
+
+    @property
+    def failed(self):
+        return not self.ok and self.code != 0
 
     def __str__(self):
         if self.ok:
             return "Build ok"
+        if self.cancelled:
+            return "Build cancelled"
```
raphendyr, line 96: Cancelled and failed both have step they are related to, thus the default print could be same (with different stateus) for both of them. On the other hand, ok result doesn't have step, so it can't be mentioned.

apluslms_roman/backends/docker.py
```diff
@@ -76,29 +76,32 @@ def _run_opts(self, task, step):
...
+                    observer.manager_msg(step, "Downloading image {}".format(step.img))
+                    try:
+                        img = client.images.pull(image, tag)
+                    except docker.errors.APIError as err:
+                        observer.step_failed(step)
+                        error = "%s %s" % (err.__class__.__name__, err)
+                        return BuildResult(step, False, -1, error)
+                observer.step_succeeded(step)
+        except KeyboardInterrupt:
+            return BuildResult(step, False)
```
raphendyr: I personally don't like to refer variables defined in for statement, outside of the for block. So I would probably wrap inside of the for block.

In all cases, there are same times when SIGTERM can be received, outside of the try-except blocks, so I guess we need one in the builder to capture those. Of course, there we won't have step info, but it kind of means that we don't know which step was the last.

Hmm.. That actually means that my code sample for result_msg needs to work bit differently. It shouldn't update steps which state is `completed`.

apluslms_roman/observer.py
raphendyr: Related to another comment, `_state_update` test should probable be change from `if self.get_step_state(step) != state:` to `cur_state = self.get_step_state(step); if cur_state != state and not cur_state.completed:`.

apluslms_roman/observer.py
```diff
@@ -151,38 +153,81 @@ def container_msg(self, step, msg):
...
 class StreamObserver(BuildObserver):
-    def __init__(self, stream=None):
+    def __init__(self, stream=None, colors=True):
         super().__init__()
         self.stream = stream or sys.stdout
+        self.colors = colors
+        self.start_time = -1
+        init_color()
+
+    def write(self, to_write, colors=None):
```
raphendyr: As this is supposed to be used internally only, maybe it should be `_write`?

apluslms_roman/observer.py
```diff
@@ -151,38 +153,81 @@ def container_msg(self, step, msg):
...
-    StepState.FAILED: "Failed!",
-    StepState.CANCELLED: "\rCancelled..",
-}
 class StreamObserver(BuildObserver):
-    def __init__(self, stream=None):
+    def __init__(self, stream=None, colors=True):
         super().__init__()
         self.stream = stream or sys.stdout
+        self.colors = colors
+        self.start_time = -1
```
raphendyr: All of these should probably be prefixed with `_` too (including the stream, which apparently was forgotten).

apluslms_roman/observer.py
ppessi: ...so should all of the steps' states be changed to `SUCCEEDED`? or can we just ignore the steps and send a result message?

apluslms_roman/backends/__init__.py
ppessi: Do you mean that the str for when a step is cancelled shoulb be "Build failed on step x"? Or "Build cancelled on step x"?

apluslms_roman/backends/docker.py
ppessi: Should I just wrap most of Builder.build in a try-except block?

apluslms_roman/backends/__init__.py
raphendyr: Second

apluslms_roman/observer.py
raphendyr: .completed is different than `SUCCEEDED`. It's any state that means step has run to the end (SUCCEEDED, FAILED, CANCELLED). Does this answer the question?

apluslms_roman/backends/docker.py
raphendyr: You could try it out at least :)

apluslms_roman/backends/__init__.py
```diff
@@ -67,21 +67,25 @@ def __str__(self):
...
+    __slots__ = ('step', 'ok', 'code', 'error')
 
-    def __init__(self, code=0, error=None, step=None):
-        self.code = code
-        self.error = error
-        self.step = step
-        assert self.ok or step is not None, "step is required for failed result"
+    def __init__(self, step=None, ok=True, code=0, error=None):
+        self.step, self.ok, self.code, self.error = (
+            step, ok, code, error)
```
raphendyr: The old syntax would be a bit more clear. In addition, I think the  `ok` should be first argument. The declaration could actually be be `def __init__(self, ok=True, *, code=0, error=None, step=None):`

apluslms_roman/backends/__init__.py
ppessi: should the observer's output have the same format? if yes, it would probably make more sense to directly use the BuildResult's str in the output

raphendyr: If I remember right, the observer loses this class before printing, so it's hard to use the same info. This str formatting probably isn't even used by cli. Possibly by GUI.

apluslms_roman/observer.py
ppessi: > Thus, what we can do (but shouldn't be mandatory), is to ensure all steps are marked (and notified) to be in SUCCEEDED state.

I was referring to that

apluslms_roman/backends/__init__.py
ppessi: It used to be used in cli (`print(result)`), but isn't anymore. The `result_msg` function could be changed to take a BuildResult as input, if that doesn't complicate things too much.

You didn't really answer wether or not you wanted me to change the output when build is cancelled in observer, too?

apluslms_roman/backends/__init__.py
ppessi: why `*`?

apluslms_roman/observer.py
raphendyr: If we end up with ok result, that means all the steps must have succeeded, thus they can be marked as such (and notified).

apluslms_roman/backends/__init__.py
raphendyr: > The result_msg function could be changed to take a BuildResult as input,

It does, but it breaks it to base types, so it could be serialized (keeps the _message protocol simple).

> wether or not you wanted me to change the output when build is cancelled in observer, too?

I think the streamobserver output is pretty much ok. It could write similar conclusion line for cancelled case too.

apluslms_roman/backends/__init__.py
raphendyr: To force keyword arguments. The order of last elements isn't very intuitive, thus forcing keywords lowers to change for mistakes.

apluslms_roman/observer.py
ppessi: With the current system, if an earlier step for some reason hasn't been marked as succeeded, the observer will print out `ok: [time]`, without really telling which step is in question (and the time will be inaccurate, too), should that be changed?

raphendyr: Add a comment about that, but it's fine as it normally doesn't happen (requires a bug). I have a plan to add serialization to streamobserver, which would actually take care of this.

raphendyr (commented):
(tried to fix github reviews state with this comment)

apluslms_roman/backends/__init__.py
```diff
@@ -67,23 +67,31 @@ def __str__(self):
...
-    __slots__ = ('code', 'error', 'step')
+    __slots__ = ('step', 'ok', 'code', 'error')
 
-    def __init__(self, code=0, error=None, step=None):
-        self.code = code
-        self.error = error
-        self.step = step
-        assert self.ok or step is not None, "step is required for failed result"
+    def __init__(self, ok=True, *, code=0, error=None, step=None):
+        self.step, self.ok, self.code, self.error = (
```
raphendyr: Can these be one per line and can we have a) assert to test that ok is false when code is != 0, or b) that ok is None by default and set to `code != 0` in that case.

apluslms_roman/backends/__init__.py
```diff
@@ -67,23 +67,31 @@ def __str__(self):
 
 
 class BuildResult:
-    __slots__ = ('code', 'error', 'step')
+    __slots__ = ('step', 'ok', 'code', 'error')
```
raphendyr: These could be one by line and sorted or in same order as in __init__.

apluslms_roman/builder.py
```diff
@@ -36,22 +36,24 @@ def build(self, step_refs: list = None, clean_build=False):
...
+                # FIXME: add support for other build paths
+                if clean_build:
+                    if isdir('_build'):
+                        rmtree('_build')
+                if clean_build or not isdir('_build'):
+                    mkdir('_build')
+                result = backend.build(task, observer)
+                observer.result_msg(result)
+            observer.done(result)
+        except KeyError:
```
raphendyr: Is this supposed to be KeyboardInterrupt?

apluslms_roman/observer.py
```diff
@@ -151,38 +154,86 @@ def container_msg(self, step, msg):
         self._send_message(Message.CONTAINER_MSG, step, msg)
 
     def result_msg(self, result):
+        state = (StepState.SUCCEEDED if result.ok
+            else (StepState.CANCELLED if result.cancelled
+                else StepState.FAILED))
+        # NOTE: if for some reason some step isn't marked as succeeded,
+        # this will result in the 'ok: [time]' message, without any
+        # indication what step is in question and with an incorrect time
```
raphendyr, line 162: With current stream observer implementation. GUIs can track incidual steps easier.

apluslms_roman/observer.py
```diff
@@ -151,38 +154,86 @@ def container_msg(self, step, msg):
...
+        if state != StepState.SUCCEEDED:
+            self._state_update(result.step, state)
```
raphendyr: The current logic, doesn't update earlier steps... I think that part is missing.

apluslms_roman/observer.py
```diff
@@ -151,38 +154,86 @@ def container_msg(self, step, msg):
         self._send_message(Message.CONTAINER_MSG, step, msg)
 
     def result_msg(self, result):
+        state = (StepState.SUCCEEDED if result.ok
+            else (StepState.CANCELLED if result.cancelled
+                else StepState.FAILED))
```
raphendyr: Can we have this in format:

```
        state = (StepState.SUCCEEDED if result.ok else
                 StepState.CANCELLED if result.cancelled else
                 StepState.FAILED)
```

I think it would be easier to read.

apluslms_roman/builder.py
ppessi: ...wonder what happened there. Probably just started writing with Key and then used autocomplete and didn't think much about it :')

apluslms_roman/observer.py
ppessi: ...right I stashed that and then forgot it

apluslms_roman/observer.py
```diff
@@ -151,38 +154,90 @@ def container_msg(self, step, msg):
...
-    StepState.CANCELLED: "\rCancelled..",
-}
 class StreamObserver(BuildObserver):
-    def __init__(self, stream=None):
+    def __init__(self, stream=None, colors=True):
         super().__init__()
-        self.stream = stream or sys.stdout
+        self._stream = stream or sys.stdout
+        self._colors = colors
+        self._start_time = -1
```
ppessi, line 176: is there a point having this here or should it be None? if the observer tries to use `_start_time` before it is set in _message for the first time, something is wrong. Another option would be setting it to the current time. (the point is to measure the time of individual steps, so that's why it isn't set to the current time at the beginning, because there are no steps)

raphendyr:
Manually merged in ff2f06f7d786b1a7df066bf87ad2cc4e48713d6d

PR #21, opened 2019-08-13
Env loading order

ppessi:
https://trello.com/c/CiMuK6FN/22-fix-backend-environment-settings-loading

### DIFF
--- a/apluslms_roman/backends/__init__.py
+++ b/apluslms_roman/backends/__init__.py
@@ -86,7 +86,7 @@ class BuildResult:
         return "Build failed on step {}: {}".format(self.step, error)
 
 
-Environment = namedtuple('Environment', [
+BackendContext = namedtuple('BackendContext', [
     'uid',
     'gid',
     'environ',
@@ -98,8 +98,8 @@ class Backend:
     WORK_PATH = '/work'
     LABEL_PREFIX = 'io.github.apluslm.roman'
 
-    def __init__(self, environment: Environment):
-        self.environment = environment
+    def __init__(self, context: BackendContext):
+        self.context = context
 
     def prepare(self, task: BuildTask, observer: BuildObserver):
         raise NotImplementedError
--- a/apluslms_roman/builder.py
+++ b/apluslms_roman/builder.py
@@ -55,26 +55,42 @@ class Builder:
         return result
 
 
+class BackendError(Exception):
+
+    def __init__(self, backend):
+        super().__init__()
+        self.backend = backend
+
+
 class Engine:
     def __init__(self, backend_class=None, settings=None):
+        backend_name = None
         if backend_class is None:
             if settings and 'backend' in settings:
                 backend_class = settings['backend']
+                if 'backends' in settings and backend_class in settings['backends']:
+                    backend_name = backend_class
+                    if 'type' in settings['backends'][backend_name]:
+                        backend_class = settings['backends'][backend_name]['type']
             else:
                 from .backends.docker import DockerBackend as backend_class
         if isinstance(backend_class, str):
             if '.' not in backend_class:
                 backend_class = BACKENDS.get(backend_class, backend_class)
-            backend_class = import_string(backend_class)
+            try:
+                backend_class = import_string(backend_class)
+            except ImportError:
+                raise BackendError(backend_class)
         self._backend_class = backend_class
 
         name = getattr(backend_class, 'name', None) or backend_class.__name__.lower()
         env_prefix = name.upper() + '_'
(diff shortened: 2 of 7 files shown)


apluslms_roman/builder.py
```diff
@@ -54,26 +54,50 @@ def build(self, step_refs: list = None, clean_build=False):
...
 
-        name = getattr(backend_class, 'name', None) or backend_class.__name__.lower()
+        name = getattr(backend_class, 'name', None) or backend_class.__name__
         env_prefix = name.upper() + '_'
-        env = {k: v for k, v in environ.items() if k.startswith(env_prefix)}
-        if settings:
-            for k, v in settings.get(name, {}).items():
-                if v is not None and v != '':
-                    env[env_prefix + k.replace('-', '_').upper()] = v
+        env = {}
```
raphendyr: While rewriting this part, we should rename this to `context` or `options` (backend options). PR #8 includes rewrite of how docker backend takes this env/context and that should be taken in consideration here. The idea is to replace some environment variables with just dictionary of options. The data is used here https://github.com/apluslms/roman/blob/7ae46077a67b94d5e80b2cd8b73d8c6932ffa285/apluslms_roman/backends/docker.py#L27. 

This would mean that when we are reading values from the environment, we would strip the `env_prefix`. PR #8 includes also some environment parsing related to this.

ppessi: Renamed `env` to `options` in some places, at least.

apluslms_roman/builder.py
```diff
@@ -84,14 +84,22 @@ def __init__(self, backend_class=None, settings=None):
...
-            env.update({env_prefix + k.replace('-', '_').upper(): v
+            options.update({prefix + k.replace('-', '_').upper(): v
                 for k, v in settings['backends'][backend_name].items()
                 if k != 'type'})
+        # environment
+        options.update({k: v for k, v in environ.items() if k.startswith(prefix)})
+        # command line
+        if prefix == 'DOCKER_' and settings and 'docker' in settings:
+            options.update({prefix + k.replace('-', '_').upper(): v
+                for k, v in settings['docker'].items()})
```
ppessi: this could be made more general with `if prefix[:-1].lower() in settings`, would that work? will all backends follow that pattern?

apluslms_roman/backends/__init__.py
```diff
@@ -98,8 +98,8 @@ class Backend:
     WORK_PATH = '/work'
     LABEL_PREFIX = 'io.github.apluslm.roman'
 
-    def __init__(self, environment: Environment):
-        self.environment = environment
+    def __init__(self, options: Environment):
+        self.options = options
```
ppessi: I kind of forgot where we ended up in the conversation, was this name okay...?

raphendyr: `self.context = context` might be good, and then the value within would be `self.context.options` and `.self.context.uid` and so on.

apluslms_roman/builder.py
raphendyr:
```
name = getattr(backend_class, 'name', None) or backend_class.__name__
prefix = name.upper() + '_'
```

Thus,

```
if name in settings
```

or actually

```
options = settings.get(name, {})
```

ppessi: or well the 'name' variable exists too, that would work better

raphendyr: There are two other changes focusing this area. The multiple config per backend changes what dict is requested from settings (name of the config), but we still want the global env values based on backend base name (e.g. backend name is `my-docker` and we get settings from `my-docker`, but we load env keys `DOCKER_`

ppessi: yeah, that's exactly how it works currently?

apluslms_roman/builder.py
```diff
@@ -55,27 +55,50 @@ def build(self, step_refs: list = None, clean_build=False):
...
+        # command line
         if settings:
-            for k, v in settings.get(name, {}).items():
-                if v is not None and v != '':
-                    env[env_prefix + k.replace('-', '_').upper()] = v
-        self._environment = Environment(getuid(), getegid(), env)
+            context.update({prefix + k.replace('-', '_').upper(): v
+                for k, v in settings.get(name, {}).items()})
+
+        self._environment = Environment(getuid(), getegid(), context)
```
raphendyr: I think the naming could be `self._backend_context = BackendContext(getuid(), getgid(), options)`

apluslms_roman/builder.py
raphendyr: Apparently

PR #22, opened 2019-08-15
Img loading

ppessi:
https://trello.com/c/e6mwPssr/35-store-image-update-time-and-only-do-updates

### DIFF
--- a/apluslms-yamlidator/apluslms_yamlidator/validator.py
+++ b/apluslms-yamlidator/apluslms_yamlidator/validator.py
@@ -214,9 +214,20 @@ def _validator_with_default_feature(validator_class):
                 instance.setdefault(property, subschema["default"])
         yield from validate_properties(validator, properties, instance, schema)
 
+    def python_type(validator, types, instance, schema):
+        if not isinstance(types, list):
+            types = [types]
+
+        if not any(instance.__class__.__name__ == type_ for type_ in types):
+            types = ["'%s'" % type_ for type_ in types]
+            yield ValidationError("{} is not an instance of Python class {}"
+                .format(instance, ', '.join(types)))
+
     new = validators.extend(validator_class,
         # redefine 'properties' validator
-        validators={"properties" : set_defaults},
+        validators={
+            "properties" : set_defaults,
+            "pythonType": python_type},
         # use collections.abc classes for array and object so Changes containers work too
         type_checker=validator_class.TYPE_CHECKER.redefine_many({
             'array': (lambda checker, instance:
--- a/apluslms_roman/backends/docker.py
+++ b/apluslms_roman/backends/docker.py
@@ -62,6 +62,10 @@ You might be able to add yourself to that group with 'sudo adduser docker'.""")
             kwargs['timeout'] = timeout
         return docker.from_env(environment=env, **kwargs)
 
+    @cached_property
+    def _cache(self):
+        return DockerCache.load()
+
     def _run_opts(self, task, step):
         env = self.environment
 
@@ -99,20 +103,42 @@ You might be able to add yourself to that group with 'sudo adduser docker'.""")
 
     def prepare(self, task, observer):
         client = self._client
+        images = self._cache.images
+        day = timedelta(days=1)
         for step in task.steps:
             observer.step_preflight(step)
             image, tag = step.img.split(':', 1)
+
+            last_update= images.get(step.img, None)
+            should_update = (not last_update or
+                datetime.now() - last_update >= day)
+
             try:
-                img = client.images.get(step.img)
+                client.images.get(step.img)
+                img_found = True
             except docker.errors.ImageNotFound:
+                img_found = False
+
+            if not img_found or should_update:
                 observer.step_running(step)
-                observer.manager_msg(step, "Downloading image {}".format(step.img))
+                if img_found:
+                    observer.manager_msg(step,
+                        ("Checking for updates for {} and "
+                        "downloading if any").format(step.img))
+                else:
+                    observer.manager_msg(step,
+                        "Downloading image {}".format(step.img))
                 try:
-                    img = client.images.pull(image, tag)
+                    client.images.pull(image, tag)
+                    images[step.img] = datetime.now()
                 except docker.errors.APIError as err:
-                    observer.step_failed(step)
-                    error = "%s %s" % (err.__class__.__name__, err)
-                    return BuildResult(-1, error, step)
+                    if not img_found:
+                        observer.step_failed(step)
+                        error = "%s %s" % (err.__class__.__name__, err)
+                        return BuildResult(-1, error, step)
+                    observer.manager_msg(step, "Couldn't download image. "
+                        "Using previously downloaded image")
+
             observer.step_succeeded(step)
         return BuildResult()
 
@@ -191,3 +217,6 @@ You might be able to add yourself to that group with 'sudo adduser docker'.""")
                         out.append("  {}: {}".format(key, val))
 
         return '\n'.join(out)
+
+    def __del__(self):
+        self._cache.save()
--- a/apluslms_roman/backends/docker.py
+++ b/apluslms_roman/backends/docker.py
@@ -6,6 +6,7 @@ from os.path import join
 import docker
 from apluslms_yamlidator.utils.decorator import cached_property
 
+from ..cache_file import CacheFile
 from ..utils.translation import _
 from . import (
     Backend,
@@ -17,6 +18,17 @@ Mount = docker.types.Mount
 
 
 logger = logging.getLogger(__name__)
+
+
+class DockerCache(CacheFile):
+    name = 'roman_docker_cache'
+    schema = name
+    version = (1, 0)
+    version_key = None
+
+    @property
+    def images(self):
+        return self.setdefault('images', {})
 
 
 @contextmanager
--- a/apluslms_roman/cache_file.py
+++ b/apluslms_roman/cache_file.py
@@ -1,0 +1,39 @@
+from abc import abstractmethod
+from os import remove
+from os.path import join
+
+from apluslms_yamlidator.document import Document
+from apluslms_yamlidator.validator import ValidationError
+
+from . import CACHE_DIR
+
+class CacheFile(Document):
+
+    @property
+    @abstractmethod
+    def name(self):
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def schema(self):
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def version(self):
+        raise NotImplementedError
+
+    @classmethod
+    def load(cls, **kwargs):
+        path = join(CACHE_DIR, '%s.yml' % (cls.name,))
+        if 'allow_missing' not in kwargs:
+            kwargs['allow_missing'] = True
+        try:
+            return super().load(path, **kwargs)
+        except ValidationError:
+            remove(path)
+            return super().load(path, **kwargs)
+
+    def validate(self, quiet=True):
+        super().validate(quiet)
--- a/apluslms_roman/schemas/roman_docker_cache-v1.0.yaml
+++ b/apluslms_roman/schemas/roman_docker_cache-v1.0.yaml
@@ -1,0 +1,13 @@
+---
+$schema: "http://json-schema.org/draft-07/schema"
+title: Roman Docker Cache
+type: object
+
+additionalProperties: false
+properties:
+  version: {}
+  images:
+    type: object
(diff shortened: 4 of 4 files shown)


apluslms-yamlidator/apluslms_yamlidator/validator.py
```diff
@@ -214,9 +215,18 @@ def set_defaults(validator, properties, instance, schema):
...
+
+        if not any(instance.__class__.__name__ == type_ for type_ in types):
+            yield ValidationError(types_msg(instance, types))
+
     new = validators.extend(validator_class,
         # redefine 'properties' validator
-        validators={"properties" : set_defaults},
+        validators={
+            "properties" : set_defaults,
+            "python_type": python_type},
```
raphendyr: prefer `pythonType` to match json schema style (only in the shema, python function is of course correct).

apluslms-yamlidator/apluslms_yamlidator/validator.py
```diff
@@ -10,6 +10,7 @@
     ValidationError,
     validators,
 )
+from jsonschema._utils import types_msg
```
raphendyr: Create local implementation, as this is import from a "private" module, which can change at any time.

apluslms_roman/backends/docker.py
```diff
@@ -79,17 +81,38 @@ def prepare(self, task, observer):
...
+                if img_found:
+                    observer.manager_msg(step,
+                        "Checking for updates for {}".format(step.img))
+                observer.manager_msg(step,
+                    "Downloading image {}".format(step.img))
                 try:
-                    img = client.images.pull(image, tag)
+                    client.images.pull(image, tag)
+                    cache.mlset('images.%s' % (step.img,), datetime.now())
+                    cache.save()
```
raphendyr: You could call `save()` one at the end of the function and not in every loop. E.g. add it to try-finally.

apluslms_roman/backends/docker.py
```diff
@@ -79,17 +81,38 @@ def prepare(self, task, observer):
         for step in task.steps:
             observer.step_preflight(step)
             image, tag = step.img.split(':', 1)
+
+            day = timedelta(days=1)
+            cache = CacheFile.load(allow_missing=True)
+            last_updated = cache.mlget('images.%s' % (step.img,), None)
```
raphendyr: Overly using the mlget feels bit complex.First, you could pass tuple `.mlget(('images', step.img), None)` or you could get the images `images = cache.setdefault('images', {})` and then use `last_updated = images.get(step.img, None)`

apluslms_roman/cache_file.py
```diff
@@ -0,0 +1,32 @@
...
+from . import CACHE_DIR
+
+class CacheFile(Document):
+    name = 'roman_cache'
+    schema = name
+    version = Version(1, 0)
+    path = join(CACHE_DIR, '%s.yml' % name)
+
+    @classmethod
+    def load(cls, **kwargs):
```
raphendyr, line 28: This is a good value for a generic solutions

apluslms_roman/cache_file.py
```diff
@@ -0,0 +1,32 @@
...
+from os.path import isdir, join
+
+from apluslms_yamlidator.document import Document
+from apluslms_yamlidator.validator import ValidationError
+from apluslms_yamlidator.utils.version import Version
+
+from . import CACHE_DIR
+
+class CacheFile(Document):
+    name = 'roman_cache'
```
raphendyr: I would create a backend specific version of this, e.g. `roman_cache_docker`.

apluslms_roman/cache_file.py
```diff
@@ -0,0 +1,32 @@
...
+    schema = name
+    version = Version(1, 0)
+    path = join(CACHE_DIR, '%s.yml' % name)
+
+    @classmethod
+    def load(cls, **kwargs):
+        if not isdir(CACHE_DIR):
```
raphendyr: Isn't the dir created by `.save()`?

apluslms_roman/backends/docker.py
```diff
@@ -79,17 +81,38 @@ def prepare(self, task, observer):
         for step in task.steps:
             observer.step_preflight(step)
             image, tag = step.img.split(':', 1)
+
+            day = timedelta(days=1)
+            cache = CacheFile.load(allow_missing=True)
```
raphendyr: You reread the file in every iteration of the loop. I think once should be sufficient. You could even load it like `self.docker` (create when requested for the first time). And then call `.save()` in `def __del__`. Though, we need to ensure `__del__` is actually called, when the cli exits.

apluslms_roman/backends/docker.py
```diff
@@ -79,17 +81,38 @@ def prepare(self, task, observer):
...
             except docker.errors.ImageNotFound:
+                img_found = False
+
+            if not img_found or last_updated >= day:
                 observer.step_running(step)
-                observer.manager_msg(step, "Downloading image {}".format(step.img))
+                if img_found:
+                    observer.manager_msg(step,
+                        "Checking for updates for {}".format(step.img))
+                observer.manager_msg(step,
```
raphendyr: I don't think we need to write both messages for update case.

apluslms_roman/cache_file.py
```diff
@@ -0,0 +1,32 @@
...
+from apluslms_yamlidator.validator import ValidationError
+from apluslms_yamlidator.utils.version import Version
+
+from . import CACHE_DIR
+
+class CacheFile(Document):
+    name = 'roman_cache'
+    schema = name
+    version = Version(1, 0)
+    path = join(CACHE_DIR, '%s.yml' % name)
```
raphendyr: You could set `version_key = None` and then this Document, shouldn't store version number in the file. As the file is removew, if the validation failes, the old version file should be automatically removed. This also means, there wouldn't be `version` in the schema.

This is ok for cache files as it's valid to remove them when not needed.

apluslms-yamlidator/apluslms_yamlidator/validator.py
ppessi: well these also use camel case: https://github.com/Julian/jsonschema/blob/master/jsonschema/_validators.py
but I guess I'll change is back to snake case to match our functions

raphendyr: Yes. That library uses non pythonic way for naming the functions, so they are same as the strings in the schema files. Pretty acceptable as those functions are not exposed outside of the library.

apluslms_roman/cache_file.py
ppessi: it does actually save the version number even if you put `version_key = None`?

apluslms_roman/backends/docker.py
ppessi: it's called when build is successfull, if there's and error or if it's interrupted so pretty safe to assume it'll always be called?

apluslms_roman/cache_file.py
```diff
@@ -0,0 +1,28 @@
...
+from apluslms_yamlidator.document import Document
+from apluslms_yamlidator.validator import ValidationError
+from apluslms_yamlidator.utils.version import Version
+
+from . import CACHE_DIR
+
+class CacheFile(Document):
+    name = 'roman_cache'
+    schema = name
+    version = Version(1, 0)
```
raphendyr: I don't think these need to be here if this is only a class that implements loader for who ever uses it. For comparison, there are classes in settings.py.

apluslms_roman/docker_cache.py
```diff
@@ -0,0 +1,9 @@
+from .cache_file import CacheFile
+
+class DockerCache(CacheFile):
+    name = 'roman_docker_cache'
+    schema = name
```
raphendyr: You should add the version here. Note that you can write it as a plain tuple or a string. I prefer tuples for this though. Document metaclass converts it to a Version object.

apluslms_roman/docker_cache.py
```diff
@@ -0,0 +1,9 @@
+from .cache_file import CacheFile
+
+class DockerCache(CacheFile):
```
raphendyr: You could have this in start of backend/docker.py. Otherwise, this would be `backends/docker/cache.py`, which would require moving `backends/docker.py` to `backends/docker/__init__,py`

apluslms_roman/backends/docker.py
```diff
@@ -99,20 +103,41 @@ def _run_opts(self, task, step):
...
         client = self._client
+        images = self._cache.images
+        day = timedelta(days=1)
         for step in task.steps:
             observer.step_preflight(step)
             image, tag = step.img.split(':', 1)
+
+            last_updated = images.get(step.img, None)
+            last_updated = (day if not last_updated
+                else datetime.now() - last_updated)
```
ppessi: if this okay or would using `last_updated and last_updated >= day` when checking be better?

apluslms_roman/backends/docker.py
```diff
@@ -99,20 +103,41 @@ def _run_opts(self, task, step):
...
+                img_found = True
             except docker.errors.ImageNotFound:
+                img_found = False
+
+            if not img_found or last_updated >= day:
                 observer.step_running(step)
-                observer.manager_msg(step, "Downloading image {}".format(step.img))
+                if img_found:
+                    observer.manager_msg(step,
+                        "Checking for updates for {}".format(step.img))
```
ppessi: Is this message accurate enough...? It kind of sounds like it'll only download the image again if there are updates but 'checking for updates' just means it downloads the image again.

apluslms_roman/backends/docker.py
```diff
@@ -99,20 +103,41 @@ def _run_opts(self, task, step):
...
+                else datetime.now() - last_updated)
+
             try:
-                img = client.images.get(step.img)
+                client.images.get(step.img)
+                img_found = True
             except docker.errors.ImageNotFound:
+                img_found = False
+
+            if not img_found or last_updated >= day:
```
ppessi: should there be a way to force docker to download the image? (like if the user knows there's an update and wants to download it) besides deleting the image from the system or editing the cache...

apluslms_roman/backends/docker.py
raphendyr: It's kind of hard to read as last_updated is first a datetime, but when it's overwritten, it becomes timedelta. Type changes are not very good for readability and a new variable name should be used.

ppessi: Okay, I can change that. I meant to ask more about the `last_updated = day if not last_updated` part, feels kind of an inelegant way to solve the problem...

apluslms_roman/backends/docker.py
raphendyr: Kind of would be good to first say `checking for updates` and then `downloading`, but that feels a lot of text (I think I commented about that before). As the process is actually `trying to get update`, I think we should indicate that we do both (check and download).

apluslms_roman/backends/docker.py
raphendyr: `roman prepare -f` or something would probably be good

ppessi: okay I'll add that to the `prepare_only` branch

apluslms_roman/backends/docker.py
ppessi: hmmm `Downloading in case of updates`? does that sound good?

apluslms_roman/backends/docker.py
raphendyr: Hmmm. Logically it kind of makes a problem. If we don't know when it was updated, we should do the update now.

ppessi: there's a `if last_updated >= day` check later, so it's always true that way, but I guess I'll check it to `last_updated and last_updated >= day`

raphendyr: Or you can calculate `should_update = not last_updated or last_updated >= day` and then `if not img_found or should_update` (or you can combine it to a long if).

apluslms_roman/backends/docker.py
raphendyr: That or `Checking for updates and downloading if any`.

apluslms_roman/backends/docker.py
ppessi: Ah that works quite well, thanks

apluslms_roman/backends/docker.py
ppessi: Is there a way to actually do that? Because currently it just downloads the image without really doing any checks.

raphendyr: No. Check and download is more human way to explain synchronisation.

PR #23, opened 2019-08-15
Add labels to docker containers and add command for deleting containers

ppessi:
https://trello.com/c/lwlcn9Hl/33-tag-name-containers-in-some-way-so-roman-can-detect-which-are-created-by-it
https://trello.com/c/ZVnzi52B/34-add-command-to-remove-failed-containers-which-are-not-cleaned-automatically


apluslms_roman/backends/docker.py
```diff
@@ -128,6 +141,11 @@ def verify(self):
         except Exception as e:
             return "{}: {}".format(e.__class__.__name__, e)
 
+    def cleanup(self):
+        containers = self._client.containers.list({'label': LABEL_PREFIX})
```
raphendyr: This should the expire date too.

apluslms_roman/backends/docker.py
```diff
@@ -17,6 +17,7 @@
 
 logger = logging.getLogger(__name__)
 
+LABEL_PREFIX = 'io.github.apluslm.roman'
```
raphendyr: This could be here: https://github.com/apluslms/roman/blob/a8df0dc2324baa6d30289b2feb44e5c9349eb622/apluslms_roman/backends/__init__.py#L98

apluslms_roman/backends/docker.py
```diff
@@ -31,6 +32,8 @@ def create_container(client, **opts):
             logger.warning("Failed to stop container %s: %s", container, err)
 
 
+
```
raphendyr: Why so much space?

apluslms_roman/cli.py
```diff
@@ -842,3 +847,8 @@ def backend_test_action(context, verbose=False):
 
 if __name__ == '__main__':
     main()
+
+
+def backend_clean_action(context):
```
raphendyr, line 850: Shouldn't be after `__name__ == '__main__'`. Instead, group with other backend actions.

apluslms_roman/backends/docker.py
ppessi: hm forgot it there after moving the function away, I guess

apluslms_roman/cli.py
ppessi: ah. that _was_ my intention, didn't notice that there :')

apluslms_roman/backends/docker.py
raphendyr: You left it here too...

PR #25, opened 2019-08-20
Validator doesn't count strings as arrays

ppessi:
Validator had been changed to accept all Sequences as arrays, and that included strings.


raphendyr:
FYI: commit message should probably start with "fix".

